PR #40, opened 2025-03-23
Complete CLI processing and add compilation framework

dspinellis:


src/uu/sed/src/script_line_provider.rs
```diff
@@ -0,0 +1,254 @@
...
+    Active {
+        index: usize,
+        reader: Box<dyn BufRead>,
+        input_name: String,
+        line_number: usize,
+    },
+    Done,
+}
+
+impl ScriptLineProvider {
```
sylvestre, line 31: this file lacks comment a bit 
maybe just explain what ScriptLineProvider is doing?

src/uu/sed/src/script_line_provider.rs
```diff
@@ -0,0 +1,254 @@
...
+                    };
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+#[cfg(test)]
+mod tests {
```
sylvestre, line 134: kudos for the tests :)

src/uu/sed/src/command.rs
```diff
@@ -1,7 +1,10 @@
 // Definitions for the compiled code data structures
 //
-// This file is part of the uutils sed package.
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2025 Diomidis Spinellis
```
sylvestre, line 4: We, in uutils, usually don't add names in the copyright in the files.

I feel that it isn't bring much in general. 
The author can be easily found in git {log, blame} and it gives the impression for new comers that the file is "owned" by someone and increases the contribution bar.

However, it isn't a big deal, so, feel free to let it as it.

src/uu/sed/src/script_line_provider.rs
dspinellis: It completely lacks comments. Not sure how it happened; I guess I found it self-explanatory, but now that I read it again I understand the problem.  I added comments as needed.

PR #57, opened 2025-04-27
Implement address compilation

dspinellis:


dspinellis:
I also added unit tests for the new 'retreat' function.

src/uu/sed/src/compiler.rs
```diff
@@ -274,11 +272,150 @@ fn compile_thread(
...
+            })
+        }
+        c if c.is_ascii_digit() => {
+            let number = parse_number(lines, line)?;
+            Ok(Address {
+                atype: AddressType::Line,
+                value: AddressValue::LineNumber(number),
+            })
+        }
+        _ => panic!("invalid context address"),
```
sylvestre, line 364: panic a bit violent in rust :)
also, you might want to add the line info to help with debugging

src/uu/sed/src/delimited_parser.rs
```diff
@@ -0,0 +1,918 @@
...
+        return None;
+    }
+
+    let char_string: String = valid_chars.into_iter().collect();
+    match u32::from_str_radix(&char_string, radix)
+        .ok()
+        .and_then(char::from_u32)
+    {
+        Some(decoded) => Some(decoded),
+        None => panic!("Unable to decode numeric character escape."),
```
sylvestre, line 78: same comment about the panic :)

src/uu/sed/src/delimited_parser.rs
```diff
@@ -0,0 +1,918 @@
...
+/// Return `None` if the result is not a valid Unicode scalar.
+fn create_control_char(x: char) -> Option<char> {
+    if !x.is_ascii() {
+        return None;
+    }
+
+    let mut c = x;
+    if c.is_ascii_lowercase() {
+        c = c.to_ascii_uppercase();
+    }
```
sylvestre, line 96: maybe just do 

```suggestion
   let c = x.to_ascii_uppercase();
```

src/uu/sed/src/script_char_provider.rs
```diff
@@ -28,6 +28,15 @@ impl ScriptCharProvider {
         }
     }
 
+    /// Retreats current position by specified number or to beginning.
+    pub fn retreat(&mut self, n: usize) {
+        if n > self.pos {
+            self.pos = 0;
+        } else {
+            self.pos -= n;
+        }
```
sylvestre, line 37:
```suggestion
         self.pos = self.pos.saturating_sub(n);
```

sylvestre:
kudos for all the tests :)

src/uu/sed/src/compiler.rs
dspinellis: It's an internal error; it shouldn't happen. The code reaches this point only when `is_address_char` has detected one of the preceding characters.

src/uu/sed/src/delimited_parser.rs
dspinellis: Same here: preceding lines have already verified that this is a valid sequence for the specified radix. The panic should only be triggered if there's a code error not a user error.

src/uu/sed/src/delimited_parser.rs
dspinellis: Thanks, fixed!

src/uu/sed/src/script_char_provider.rs
dspinellis: Didn't know that; thanks! Fixed.

dspinellis:
I pushed a fix addressing the comments that needed fixing.

PR #77, opened 2025-05-19
Complete implementation of all commands, in-place editing, and optimizations

dspinellis:
At this state _sed_ implements all POSIX commands, and can correctly run the two complex scripts: [hanoi.sed](https://github.com/uutils/sed/blob/main/tests/fixtures/sed/script/hanoi.sed) (solves the Towers of Hanoi puzzle) and  [math.sed](https://github.com/uutils/sed/blob/main/tests/fixtures/sed/script/math.sed)  (implements an arbitrary precision integer math calculator).

The performance of this Rust implementation is now better than the GNU and FreeBSD implementations for most benchmark cases.

## GNU sed under Linux
```
               no-op-short rust is  3.04 times faster than gnu
          access-log-no-op rust is  1.91 times faster than gnu
       access-log-no-subst rust is  1.59 times faster than gnu
          access-log-subst rust is  1.32 times faster than gnu
         access-log-no-del rust is  1.54 times faster than gnu
        access-log-all-del rust is  1.99 times faster than gnu
       access-log-translit rust is 12.67 times faster than gnu
    access-log-complex-sub gnu is  4.24 times faster than rust
                 remove-cr rust is  1.68 times faster than gnu
              genome-subst rust is  1.15 times faster than gnu
                number-fix rust is  1.06 times faster than gnu
               long-script gnu is  2.19 times faster than rust
                     hanoi rust is  1.96 times faster than gnu
                 factorial rust is  1.25 times faster than gnu
```

## FreeBSD sed
```
               no-op-short rust is  3.50 times faster than fbsd
          access-log-no-op rust is  2.45 times faster than fbsd
       access-log-no-subst rust is  1.38 times faster than fbsd
          access-log-subst rust is  4.02 times faster than fbsd
         access-log-no-del rust is  1.35 times faster than fbsd
        access-log-all-del rust is  4.74 times faster than fbsd
       access-log-translit fbsd is  1.14 times faster than rust
    access-log-complex-sub rust is  2.00 times faster than fbsd
                 remove-cr rust is  2.48 times faster than fbsd
              genome-subst rust is  2.21 times faster than fbsd
                number-fix fbsd is  1.04 times faster than rust
               long-script fbsd is 15.46 times faster than rust
                     hanoi rust is  9.33 times faster than fbsd
                 factorial rust is 115.72 times faster than fbsd
```

## Remaining tasks
* [ ] Improve runtime error reporting by including script coordinates in each command.
* [ ] Fix buffering on terminal output to match current implementations
* [ ] Profile lagging benchmarks to see whether there's room for further optimizations.
* [ ] Implement more GNU extensions.


src/uu/sed/src/processor.rs
```diff
@@ -161,6 +166,21 @@ fn write_chunk(
     Ok(())
 }
 
+/// Return the RE or the saved RE if the RE is None.
+/// Update the saved RE to RE.
+fn re_or_saved_re(regex: &Option<Regex>, context: &mut ProcessingContext) -> UResult<Regex> {
+    match regex {
+        Some(re) => {
+            *context.saved_regex.borrow_mut() = Some(re.clone());
```
sylvestre, line 174: would it be possible to ignore the clone ?

src/uu/sed/src/processor.rs
```diff
@@ -175,8 +195,20 @@ fn substitute(
...
         count += 1;
+        let caps = match caps {
+            Ok(c) => c,
+            Err(e) => {
+                return Err(USimpleError::new(
+                    2,
+                    format!("regular expression capture retrieval error: {}", e),
+                ));
+            }
+        };
```
sylvestre, line 210: as you wish but you could write:

```suggestion
     let caps = caps
    .map_err(|e| USimpleError::new(
        2,
        format!("regular expression capture retrieval error: {}", e),
    ))?;

```

dspinellis: Thanks, fixed!

src/uu/sed/src/processor.rs
dspinellis: Indeed; fixed.

sylvestre: impressive!

src/uu/sed/src/fast_io.rs
```diff
@@ -469,6 +469,16 @@ impl OutputBuffer {
             false,
         )))
     }
+
+    /// Copy the specified file to the output.
+    pub fn copy_file(&mut self, path: &PathBuf) -> io::Result<()> {
+        #[cfg(unix)]
```
sylvestre, line 475: do you know what will happen here on !unix ?

dspinellis:
@sylvestre I integrated the commits for the remaining commands and in-place editing. The implementation is mostly done now.

sylvestre:
I will reply to your email too :)

PR #108, opened 2025-07-22
Refactor command compilation to use function pointers

dspinellis:
This simplifies command handling by
* replacing the construction and a runtime map lookup with a simple `match` structure, 
* eliminating an extra `match` lookup on the command argument type, 
* eliminating one field in each command specification, and 
* making command processing similar for each command.


cakebaker:
Nice refactoring, thanks!

PR #131, opened 2025-08-19
Reorder s/// struct elements for readability

dspinellis:
Make their order match how they are specified.
While at it remove unused line_number field.


cakebaker:
Thanks!

PR #161, opened 2025-09-20
Use copy_file_range(2) for improved performance where possible

dspinellis:


dspinellis:
@sylvestre yes, ready, thanks!

PR #197, opened 2025-11-23
 rename uu_sed to sed 

sylvestre:
for publication

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,6 @@ once_cell = { workspace = true }
 phf = { workspace = true }
 predicates = { workspace = true }
 regex = { workspace = true }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -67,8 +66,10 @@ memmap2.workspace = true
 once_cell = { workspace = true }
 phf = { workspace = true }
 predicates = { workspace = true }
-sed = { optional = true, version = "0.0.1", package = "uu_sed", path = "src/uu/sed" }
+regex = { workspace = true }
+# sed module is now part of the main crate
```
cakebaker: While the comment makes sense in the context of this PR, I think it is not that helpful when you look at this file at a later time. 
```suggestion
```

PR #213, opened 2025-12-14
Fixes newline

### LINKED ISSUE #143: Missing newline in output with `-e p`
While trying out the fuzzer I noticed that GNU `sed` adds a newline in the output if the input doesn't end with a newline. Our implementation doesn't.
```
$ printf "foo" | sed -e p
foo
foo$

$ printf "foo" | cargo run -q sed -e p
foofoo$
```

pocopepe:
Closes #143 

This fixes the `p` command so it matches GNU sed’s `outputline()` behavior.

`write_chunk()` now writes the line content without a trailing newline, and only appends `\n` if the original input line was newline-terminated (GNU’s `chomped = true` case). With this in place, the `p` command can simply call `write_chunk()` directly without any special handling.

To fully mirror GNU sed, this newline logic would ideally apply to all `write_chunk()` calls. However, doing so changes existing output behavior and breaks a large number of tests that currently depend on the old semantics. For now, this change is intentionally scoped to `p` to fix the reported issue without introducing broader regressions.

### DIFF
--- a/tests/by-util/test_sed.rs
+++ b/tests/by-util/test_sed.rs
@@ -1117,3 +1117,14 @@ fn test_missing_address_re() {
         .code_is(2)
         .stderr_is("sed: <script argument 1>:2:3: 'input/lines1':1 error: no previous regular expression\n");
 }
+
+////////////////////////////////////////////////////////////
+// Test for issue #143: Missing newline in output with `-e p`
+#[test]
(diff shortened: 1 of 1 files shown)


sylvestre:
added the test here: https://github.com/uutils/sed/commit/35b6e84c1616b88a6b2ce1159c9459b42935e1af

PR #247, opened 2026-01-11
WIP: Publish comments in case of changes with the GNU test suite

sylvestre:


sylvestre:
purely ci, don't impact the product, automerging

PR #256, opened 2026-01-11
fix a bug with the deletion pattern

sylvestre:


dspinellis (commented):
Rather than adding and processing another flag, can we perhaps achieve the same by adjusting the `IOChunk` `has_newline` field?

dspinellis:
Moved the tests to https://github.com/uutils/sed/pull/535. They seem to pass now.

PR #259, opened 2026-01-12
Preserve new line on exchanges (x) with empty hold spaces.

### LINKED ISSUE #254: handles exchange command (x) with empty hold space differently from GNU (missing  newline)
Rust (incorrect - missing newline)
```
  $ printf "a\nb\nc" | cargo run -p sed -- '2x'
  a
  c
  # Exit code: 0

```

  GNU (correct)
```
  $ printf "a\nb\nc" | /usr/bin/sed '2x'
  a

  c
  # Exit code: 0
```

AlexanderWangY:
Fixes #254 

- On exchange where hold space is empty, we ignore exchanging the newline.
- Only when the hold space is filled or the hold space has a newline do we exchange the newline
- Added test case

> Note: It got me thinking, will there ever be a case where we exchange a line and have the new space be part of the line itself? It seems as if we won't really have to set it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -505,7 +505,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


src/sed/processor.rs
```diff
@@ -631,8 +631,10 @@ fn process_file(
                 'x' => {
                     // Exchange the contents of the pattern and hold spaces.
                     let (pat_content, pat_has_newline) = pattern.fields_mut()?;
+                    if !context.hold.content.is_empty() || context.hold.has_newline {
+                        std::mem::swap(pat_has_newline, &mut context.hold.has_newline);
```
sylvestre: please add a comment explaining what it is doing

sylvestre: ping ? :)

AlexanderWangY: hii! so sorry about pushing this off.... I've made the requested changes!

Issue #253, opened 2026-01-11
handles multiple exchange commands differently from GNU (pattern space  truncation)


sylvestre:
Rust (incorrect - pattern space truncated)
```
  $ echo -e "line1\nline2\nline3" | cargo run -p sed -- '2x;3x'
  line1
  line2
  # Exit code: 0

```

  GNU (correct)
 ```
 $ echo -e "line1\nline2\nline3" | /usr/bin/sed '2x;3x'
  line1

  line2
  # Exit code: 0

```

PR #262, opened 2026-01-14
Fix truncated pattern space problem

LoukasPap:
Fix [issue 253](https://github.com/uutils/sed/issues/253)

- Initialize the context of the hold space to `\n` (newline character)
- Previously, it started as an empty string which resulted to wrong outputs

This solution not only fixes the problem for the swap pattern `x` but also for `g` and `G` .


LoukasPap:
Creating a new PR with only the commits needed.

PR #267, opened 2026-01-18
Improve command range handling

dspinellis:
Before this commit, the output of sed was lagging one line behind, as it was always checking whether each line was the last one ($).  This change modifies the behavior to perform this check only when actually needed. This brings sed's interactive behavior (when reading from a terminal) in line with the GNU and FreeBSD implementations.

In addition this PR adds support for the GNU address 0 extension and for ~step ranges. To support this it refactors the address representation.


dspinellis:
On it.

PR #311, opened 2026-02-13
Use rustc-hash

oech3:
+2~-1% perf. Did not help dropping cgu. Interested in mimalloc+this one.


sylvestre:
@oech3 what is the plan here? thanks :)

PR #317, opened 2026-02-19
reproducible fuzzing, fast setup

oech3:


cakebaker:
I don't know if you have seen it: the `Run fuzz_sed for XX seconds` step fails with the following compilation error:
```
error: sanitizer is incompatible with statically linked libc, disable it using `-C target-feature=-crt-static`

error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-unknown-linux-musl` target may not be installed
  = help: consider downloading the target with `rustup target add x86_64-unknown-linux-musl`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`
```

PR #328, opened 2026-03-04
Add validations for brace quantifiers {m}, {m,} and {m,n}

### LINKED ISSUE #289: ERE quantifiers are not validated: `{m}`, `{m,}` and `{m,n}`
revision: 35041093732572f30400b5145a158b2b8cdae6e0

https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap09.html#tag_09_04_06

> When an ERE matching a single character or an ERE enclosed in parentheses is followed by an interval expression of the format <tt>"{m}"</tt>, <tt>"{m,}"</tt>, or <tt>"{m,n}"</tt>, together with that interval expression it shall match what repeated consecutive occurrences of the ERE would match. The values of <i>m</i> and <i>n</i> are decimal integers in the range 0&lt;=<i>m</i>&lt;=<i>n</i>&lt;={RE_DUP_MAX}, where <i>m</i> specifies the exact or minimum number of occurrences and <i>n</i> specifies the maximum number of occurrences. The expression <tt>"{m}"</tt> matches exactly <i>m</i> occurrences of the preceding ERE, <tt>"{m,}"</tt> matches at least <i>m</i> occurrences, and <tt>"{m,n}"</tt> matches any number of occurrences between <i>m</i> and <i>n</i>, inclusive.

But uutils does not check the condition—*m*&gt;*n*:

```console
$ gnused -n -E -e '/x{2,1}/p' <<< x
sed: -e expression #1, char 8: Invalid content of \{\}
$ echo $?
1
$ bsdsed -n -E -e '/x{2,1}/p' <<< x
bsdsed: 1: "/x{2,1}/p
": RE error: Invalid content of \{\}
$ echo $?
1
$ cargo run --quiet -- -n -E -e '/x{2,1}/p' <<< x
x
$ echo $?
0
```

and negative *m*, e.g. `x{-1}` and `x{-1,}`.

LoukasPap:
Fixes [issue 289](https://github.com/uutils/sed/issues/289)

There are two types of errors in quantifiers:
1. `Unmatched \{` for unterminated braces
2. `Regular expression too big` if m and/or n are outside the [0,255] bounds
3. `Invalid content of \{\}` for all the others

### `Unmatched \{`
`sed` always captures the unmatched braces error first. To deal with that, I scan ahead the regex for unclosed `{ }` and if it is violated it throws the error, otherwise we retreat back to the beginning to start the processing.

### `Regular expression too big`
- [x] m and/or n must be between [0, 32767]

### `Invalid content of \{\}`
It checks for:
- [x] m <= n
- [x] integers and one comma are only allowed (the comma between m and n)
- [x] `{,}` to be converted to `0,`
- [x] `{,n}` to be converted to `{0,n}`

### DIFF
--- a/src/sed/compiler.rs
+++ b/src/sed/compiler.rs
@@ -12,7 +12,9 @@ use crate::sed::command::{
     Address, Command, CommandData, ProcessingContext, ReplacementPart, ReplacementTemplate,
     Substitution, Transliteration,
 };
-use crate::sed::delimited_parser::{parse_char_escape, parse_regex, parse_transliteration};
+use crate::sed::delimited_parser::{
+    RegexMode, parse_char_escape, parse_regex, parse_transliteration,
+};
 use crate::sed::error_handling::{ScriptLocation, compilation_error, semantic_error};
 use crate::sed::fast_regex::Regex;
 use crate::sed::named_writer::NamedWriter;
@@ -286,7 +288,6 @@ fn compile_sequence(
         let n_addr = compile_address_range(lines, line, &mut cmd, context)?;
         line.eat_spaces();
         let mut cmd_spec = get_verified_cmd_spec(lines, line, n_addr, context.posix)?;
-
         // Compile the command according to its specification.
         let mut cmd_mut = cmd.borrow_mut();
         cmd_mut.code = line.current();
@@ -331,10 +332,8 @@ fn compile_address_range(
     let mut is_line0 = false;
 
     line.eat_spaces();
-    if !line.eol()
-        && is_address_char(line.current())
-        && let Ok(addr1) = compile_address(lines, line, context)
-    {
+    if !line.eol() && is_address_char(line.current()) {
+        let addr1 = compile_address(lines, line, context)?;
         is_line0 = matches!(addr1, Address::Line(0));
         cmd.addr1 = Some(addr1);
         if is_line0 && context.posix {
@@ -364,9 +363,8 @@ fn compile_address_range(
         }
 
         // Look for second address.
-        if !line.eol()
-            && let Ok(addr2) = compile_address(lines, line, context)
-        {
+        if !line.eol() {
+            let addr2 = compile_address(lines, line, context)?;
             // Set step_n to the number specified in the (required numeric) address.
             let step_n = if is_step_match || is_step_end {
                 match addr2 {
@@ -449,7 +447,12 @@ fn compile_address(
                 // The next character is an arbitrary delimiter
                 line.advance();
             }
-            let re = parse_regex(lines, line)?;
+            let regex_mode = if context.regex_extended {
+                RegexMode::Extended
+            } else {
+                RegexMode::Basic
+            };
+            let re = parse_regex(lines, line, regex_mode)?;
             // Skip over delimiter
             line.advance();
 
@@ -532,7 +535,7 @@ fn parse_command_ending(
 }
 
 /// Convert a primitive BRE pattern to a safe ERE-compatible pattern string.
-/// - Replaces `\(`, `\)`, `\?`, `\+` and `\|` with `(`, `)`, `?`, `+` and `|`.
+/// - Replaces `\(`, `\)`, `\?`, `\+`, `\|`, `\{` and `\}` with `(`, `)`, `?`, `+`, `|`, `{` and `}`.
 /// - Puts single-digit back-references in non-capturing groups..
 /// - Escapes ERE-only metacharacters: `+ ? { } | ( )`.
 /// - Leaves all other characters as-is.
@@ -564,6 +567,14 @@ fn bre_to_ere(pattern: &str) -> String {
                 Some('|') => {
                     chars.next();
                     result.push('|'); // Alternation operator
+                }
+                Some('{') => {
+                    chars.next();
+                    result.push('{'); // Brace quantifier start
+                }
+                Some('}') => {
+                    chars.next();
+                    result.push('}'); // Brace quantifier end
                 }
                 Some(v) if v.is_ascii_digit() => {
                     // Back-reference.  In sed BREs these are single-digit
@@ -635,7 +646,7 @@ fn compile_regex(
 
     // Convert basic to extended regular expression if needed.
     let pattern = if context.regex_extended {
-        pattern
+        &pattern.replace("{,}", "*")
     } else {
         &bre_to_ere(pattern)
     };
@@ -644,7 +655,7 @@ fn compile_regex(
     let pattern = if icase {
         format!("(?i){pattern}")
     } else {
-        pattern.to_string()
+        pattern.clone()
     };
 
     // Compile into engine.
@@ -786,8 +797,12 @@ fn compile_subst_command(
         );
     }
 
-    let pattern = parse_regex(lines, line)?;
-
+    let regex_mode = if context.regex_extended {
+        RegexMode::Extended
+    } else {
+        RegexMode::Basic
+    };
+    let pattern = parse_regex(lines, line, regex_mode)?;
     let mut subst = Box::new(Substitution::default());
 
     subst.replacement = compile_replacement(lines, line)?;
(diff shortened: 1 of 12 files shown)


e-kwsm (commented):
> {,n} to be parsed as literal

nitpick: `"{,n}"` pattern is not specified in ERE, but

```bash
<<< xxx sed -E -n -e '/x{,-1}/s//@/p'
```

- GNU and BSD: `Invalid content of \{\}`, exit with one
- uutils: prints nothing and exits with zero

```bash
<<< xxx sed -E -n -e '/x{,0}/s//@/p'
```

- GNU and BSD: print `@xxx`
- uutils: prints nothing

```bash
<<< xxx sed -E -n -e '/x{,1}/s//@/p'
```

- GNU and BSD: print `@xx`
- uutils: prints nothing

e-kwsm:
```bash
<<< xxx sed -n -E -e '/.{,x}/p'
```

- GNU and BSD: `Invalid content of \{\}`
- uutils: prints nothing

e-kwsm (approved):

e-kwsm (approved):

src/sed/delimited_parser.rs
```diff
@@ -756,53 +951,97 @@ mod tests {
...
-        let parsed = parse_regex(&lines, &mut line).unwrap();
+        let parsed = parse_regex(&lines, &mut line, false).unwrap();
         assert_eq!(parsed, "ab\n");
         assert_eq!(line.current(), '/');
     }
 
+    #[test]
+    fn test_basic_regex_quantifier() {
+        let (lines, mut line) = make_providers("/a\\{2,3\\}/p");
+        let parsed = parse_regex(&lines, &mut line, true).unwrap();
```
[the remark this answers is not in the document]
LoukasPap: Correct, my bad

src/sed/delimited_parser.rs
```diff
@@ -353,6 +384,170 @@ pub fn parse_regex(lines: &ScriptLineProvider, line: &mut ScriptCharProvider) ->
     compilation_error(lines, line, "unterminated regular expression")
 }
 
+// Check for closing brace and the structure/content.
+fn validate_quantifier_structure(
```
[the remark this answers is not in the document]
LoukasPap: I could add a function in the ScriptCharProvider impl that sets the line to the specified position, which in our case we want it to be the starting position of the quantifier (the `{`).

src/sed/delimited_parser.rs
```diff
@@ -312,11 +312,15 @@ fn scan_delimiter(lines: &ScriptLineProvider, line: &mut ScriptCharProvider) ->
...
 /// Parse the regular expression delimited by the current line
 /// character and return it as a string.
-/// On return the line is on the closing delimiter.
-pub fn parse_regex(lines: &ScriptLineProvider, line: &mut ScriptCharProvider) -> UResult<String> {
+/// On return, the line is on the closing delimiter.
+/// If extended_mode is false, quantifiers like {m,n} are treated as literals.
+pub fn parse_regex(
+    lines: &ScriptLineProvider,
+    line: &mut ScriptCharProvider,
+    is_extended_mode: bool,
```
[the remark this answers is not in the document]
LoukasPap: I am thinking of using an enum to represent if the mode is Basic or Extended.

LoukasPap:
I will make changes in the next few days and re-push.

LoukasPap:
Just rebased atop main and squashed all commits into one. I am open to new approaches and suggestions for a better solving of the problem, if it's needed :)

src/sed/delimited_parser.rs
```diff
@@ -15,6 +15,14 @@ use crate::sed::script_line_provider::ScriptLineProvider;
 use std::char;
 use uucore::error::UResult;
 
+/// Defines whether regex patterns use Basic Regular Expression (BRE) or
+/// Extended Regular Expression (ERE) syntax.
+#[derive(Copy, Clone, Debug)]
+pub enum RegexMode {
+    Basic,
+    Extended,
+}
```
LoukasPap: Now that I see this, it would better be located inside `command.rs`.

src/sed/delimited_parser.rs
```diff
@@ -353,6 +393,173 @@ pub fn parse_regex(lines: &ScriptLineProvider, line: &mut ScriptCharProvider) ->
...
+    }
+    // Parse m value
+    let mut m = String::new();
+    while line.current() != ',' && line.current() != '}' && line.current() != '\\' {
+        m.push(line.current());
+        line.advance();
+    }
+    let m_val: u32 = match m.parse() {
+        Ok(val) => {
+            if val > 255 {
```
sylvestre: Upper bound is wrong — 255 should be 32767 (RE_DUP_MAX). no ?

tests/fixtures/sed/output/bre_quantifier_comma
```diff
@@ -0,0 +1,5 @@
+Hello World
```
sylvestre: please generate them on the fly

src/sed/compiler.rs
```diff
@@ -636,7 +647,7 @@ fn compile_regex(
 
     // Convert basic to extended regular expression if needed.
     let pattern = if context.regex_extended {
-        pattern
+        &pattern.replace("{,}", "*")
```
sylvestre: are you sure ?

  $ printf 'x{,}y\n' | sed   -E 's/[{,}]/_/g' 
should return
x___y

src/sed/compiler.rs
```diff
@@ -331,10 +332,8 @@ fn compile_address_range(
     let mut is_line0 = false;
 
     line.eat_spaces();
-    if !line.eol()
-        && is_address_char(line.current())
-        && let Ok(addr1) = compile_address(lines, line, context)
-    {
+    if !line.eol() && is_address_char(line.current()) {
```
sylvestre, line 333: is it related?

src/sed/delimited_parser.rs
```diff
@@ -353,6 +393,173 @@ pub fn parse_regex(lines: &ScriptLineProvider, line: &mut ScriptCharProvider) ->
...
+
+    if invalid_content_detected {
+        return compilation_error(lines, line, invalid_content_error_msg);
+    }
+
+    line.set_position(initial_pos);
+    Ok(initial_pos)
+}
+
+// Peforms validations on m and/or n values of the quantifier
```
sylvestre:
```suggestion
// Performs validations on m and/or n values of the quantifier
```

src/sed/compiler.rs
LoukasPap: If we keep it as it was, errors are being silently skipped without back-propagating them. Moving `compile_address(lines, line, context)?;` in a separate line catches compilation errors. Or did you mean something else?

tests/fixtures/sed/output/bre_quantifier_comma
LoukasPap: Sorry, forgot to remove the output files

src/sed/delimited_parser.rs
LoukasPap: I had seen [here](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html#:~:text=048-,%7B%5FPOSIX2%5FRE%5FDUP%5FMAX,255) that the value is 255, but 32767 is the upper limit as I tested it for both GNU and POSIX, so you are right.

src/sed/compiler.rs
LoukasPap: Yes, it returns x___y.
I can change the place where replacing happens to somewhere it makes more sense, if it is necessary.

src/sed/delimited_parser.rs
```diff
@@ -353,6 +393,173 @@ pub fn parse_regex(lines: &ScriptLineProvider, line: &mut ScriptCharProvider) ->
...
+    lines: &ScriptLineProvider,
+    line: &mut ScriptCharProvider,
+) -> UResult<String> {
+    line.advance();
+
+    // Handle {,} (zero or more) special case
+    if line.current() == ',' {
+        line.advance();
+        if line.current() == '}' {
+            return Ok(",".to_string());
```
LoukasPap: @sylvestre I can change `*` to `0,` to avoid replacement later

PR #345, opened 2026-03-19
Improve GNU sed testsuite compatibility

kevinburkesegment:
- Fix `l` command default width from 60 to 70 (matching GNU sed)
- Wire `-l` flag and COLS env var to `l` command width
- Reject `l<N>` (GNU extension) in --posix mode
- Accept `}` and `#` as valid command terminators (fixes command-endings)
- Accept `}` and `#` as valid substitute flag terminators
- Eat spaces before command endings (e.g. `y/1/a/ ;d`)
- Scope `#n` silent mode to first script source only (fixes comment-n)
- Drop backslash for unknown escape sequences in s/// replacement (GNU behavior)
- Match GNU sed error message format: `-e expression #N, char C:` for string scripts, `file:line:` for file scripts
- Add `get_source_index()` to ScriptLineProvider
- Add test-results.txt to .gitignore

GNU testsuite results: 7 → 9 passing (command-endings, comment-n now pass).

### DIFF
--- a/src/sed/compiler.rs
+++ b/src/sed/compiler.rs
@@ -27,7 +27,9 @@ use std::rc::Rc;
 use terminal_size::{Width, terminal_size};
 use uucore::error::{UResult, USimpleError};
 
-const DEFAULT_OUTPUT_WIDTH: usize = 60;
+/// Default line-wrap width for the `l` command when no terminal, COLS, or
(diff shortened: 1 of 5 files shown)


sylvestre:
@kevinburkesegment could you please rebase it? thanks

PR #369, opened 2026-04-12
Cargo.toml: bump Rust version

oech3:
Since this project is WIP and none of (lts) distribution package it, there is no reason to restrict toolchain at here and build time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ readme = "README.md"
 keywords = ["sed", "uutils", "cross-platform", "cli", "utility"]
 categories = ["command-line-utilities"]
 edition = "2024"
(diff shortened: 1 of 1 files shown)


xtqqczze:
This crate (unlike https://github.com/uutils/tar/pull/259) has already been released to crates.io, so we should be a little more cautious about MSRV.

It’s also worth noting that, unlike a `rust-toolchain` file, the `rust-version` field does not restrict the use of a newer toolchain.

sylvestre:
i don't think we should do it yet. 
let's bump when we have a constraint

xtqqczze:
@oech3 did you have any features in mind you think we could use?

Issue #266, opened 2026-01-18
`g` and `G` discrepancy when the patten space is not set


e-kwsm:
c6479b44ad05b71fd699b059ae1e32fa69e738aa

```console
$ seq 3 | cargo run --quiet -- -e '2g'
1
3
$ seq 3 | cargo run --quiet -- -e '2G'
1
2
3
```

GNU and BSD sed&rsquo;s:

```console
$ seq 3 | sed -e '2g'
1

3
$ seq 3 | sed -e '2G'
1
2

3
```

[POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/sed.html):

> [2addr]g
>     Replace the contents of the pattern space by the contents of the hold space.
> [2addr]G
>     Append to the pattern space a <newline> followed by the contents of the hold space.

If the hold space is not null e.g. `1h;2g`, all of them behave in the same way.

LoukasPap:
@e-kwsm This is solved in [PR 263](https://github.com/uutils/sed/pull/263).

PR #378, opened 2026-04-21
fix: correct newline handling for p and x commands

### LINKED ISSUE #306: `1x` should output empty line if input doesn't contain newline
GNU `sed` outputs an empty line with `1x` if the input doesn't contain a newline. uutils `sed`, on the other hand, doesn't:
```bash
$ printf "abc" | sed 1x

$ 
$ printf "abc" | cargo run -q sed 1x
$
```
The exception is an empty string as input: in such a case there is no output.
```bash
$ printf "" | sed 1x
$ 
$ printf "" | cargo run -q sed 1x
$
```

pocopepe:
Fixes #306 and #143

- Hold space was initialized without `has_newline: true`, causing `1x` to output nothing instead of an empty line when input has no trailing newline
- `p` was unconditionally appending `\n`, breaking `-n p` on no-newline input
 
Fixed by initializing hold space as newline terminated and adding a `pending_newline` mechanism in `OutputBuffer` that defers newline injection until the next write, so `-n` mode never triggers it
Also incidentally fixes the `g`/`G` empty hold space behavior from #266 (overlaps with open PR #263 same bug, different approach)

### DIFF
--- a/src/sed/fast_io.rs
+++ b/src/sed/fast_io.rs
@@ -559,8 +559,11 @@ pub struct OutputBuffer {
     max_pending_write: usize,       // Max bytes to keep before flushing
     #[cfg(unix)]
     mmap_chunk: Option<MmapOutput>, // Chunk to write
+    // True when the last write didn't end with \n; the \n is deferred so
+    // that commands like `p` don't emit a spurious newline under -n.
+    pending_newline: bool,
     #[cfg(test)]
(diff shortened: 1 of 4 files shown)


dspinellis (approved):
LGTM. Please squash the commits into a single one.

pocopepe:
squashed into one commit :)

PR #379, opened 2026-04-21
assert_fs belongs in dev-dependencies

pocopepe:
Closes #323 
A simple move to fix the wasm build


cakebaker:
Good catch, thanks!

PR #409, opened 2026-05-03
Should show usage without argments #274

### LINKED ISSUE #274: Should show usage without argments
```
$ gnu-sed
Usage: sed [OPTION]... {script-only-if-no-other-script} [input-file]...

  -n, --quiet, --silent
```
But https://github.com/uutils/sed/releases/tag/latest-commit does:
```
sed: missing script
Try sed --help' for more information.
```

Devel08:
resolves #274

### DIFF
--- a/tests/by-util/test_sed.rs
+++ b/tests/by-util/test_sed.rs
@@ -54,7 +54,7 @@ fn test_silent_alias() {
 #[test]
 fn test_missing_script_argument() {
     new_ucmd!()
-        .args(&["-i", "file.txt"])
+        .args(&["-n"])
         .fails()
         .code_is(1)
         .stderr_contains("missing script");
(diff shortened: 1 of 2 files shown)


sylvestre:
Please add a test to make sure we don't regress in the future? Thanks

Devel08:
@sylvestre Done, I also changed the test_missing_script_argument test, because it inputted no arguments, thus it was expected to receive --help. It now inputs arguments only for a file, but not a script

Devel08:
I always forget the cargo fmt :/

tests/by-util/test_sed.rs
```diff
@@ -54,11 +54,20 @@ fn test_silent_alias() {
 #[test]
 fn test_missing_script_argument() {
     new_ucmd!()
+        .args(&["-i", "file.txt"])
```
dspinellis: The argument doesn't match the option's semantics. Consider `.backup` or just passing the `-n` flag.

dspinellis (changes requested):
Thanks! One minor improvement and we're good to go.

Issue #404, opened 2026-04-30
Add `e` substitute flag (execute replacement as shell)


sylvestre:
## Bug

The `e` substitute flag (execute the result of the substitution as a
shell command, replace pattern space with the command's stdout) is
not implemented. It is the substitute-time counterpart to the standalone
`e` command (which is also missing — together they would unblock the
`eval` and `sandbox` GNU tests).

## Reproduction

```sh
$ echo a | /usr/bin/sed 's/.*/echo hi/e'
hi

$ echo a | ./target/release/sed 's/.*/echo hi/e'
sed: <script argument 1>:1:14: error: invalid substitute flag: 'e'
```

## What it should do

GNU manual:

> `e`  This command allows one to pipe input from a shell command into
> the pattern space. Without parameters, the e command executes the
> command that is found in pattern space and replaces the pattern
> space with the output; a trailing newline is suppressed. […] As a
> GNU extension, the s command can have an e modifier with the same
> functionality.

Order matters: the substitution is performed first, then the rendered
replacement is run as a shell command, and its stdout (sans trailing
newline) replaces the pattern space.

Must be rejected under `--sandbox` (already plumbed at `mod.rs:104`)
and under `--posix`.

## Suspected place to add it

`src/sed/compiler.rs:854` — `compile_subst_flags`. Add an `'e' =>` arm
that sets `subst.execute = true` (new field), and reject the flag when
either `posix` or `sandbox` is true.

The execution side belongs in `processor.rs`, after the substitution
result is rendered: spawn a shell (`/bin/sh -c <result>`), read its
stdout, replace the pattern space.

## Security note

`s///e` and the standalone `e` command are why `--sandbox` exists.
Make sure the sandbox check is enforced at compile time (so the
script is rejected before any input is read), not only at run time.

## Affected GNU testsuite tests

`subst-options`, `sandbox`.

PR #413, opened 2026-05-06
Add `e` substitution flag

PranavRJoshi:
This PR introduces `e` substitution mentioned in #404.

## What this does
Implements the `e` substitute flag, which executes the result of the substitution as a shell command and replaces the pattern space with stdout (minus trailing newline).

```shell
$ echo a | ./target/release/sed  's/.*/echo hi/e'
hi

$ echo a | ./target/release/sed --posix 's/.*/echo hi/e'
sed: <script argument 1>:1:14: error: the 'e' substitute flag is not allowed with --posix or --sandbox
```

## Changes
- `src/sed/command.rs`: new `execute: bool` field on `Substitution`
- `src/sed/compiler.rs`: `'e'` arm in `compile_subst_flags`; rejected at compile time under `--posix` or `--sandbox`.
- `src/sed/processor.rs`: shell execution via `/bin/sh -c` after substitution

## Tests
- Three new unit tests for the flag parser covering the intended path, --posix rejection, and --sandbox rejection.
- Full `cargo test` run passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -392,9 +392,9 @@ dependencies = [
 
 [[package]]
 name = "ctor"
-version = "1.0.4"
+version = "1.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e2049d5ec0c16ffe3815be58fe6966159a0f37dcf16ae1c7330808ed7c9b71a1"
+checksum = "378f0974ae2468eaf63aa036dbe9c926b0dc7ea64c156f2ea618bc2f75b934f0"
 dependencies = [
  "link-section",
  "linktime-proc-macro",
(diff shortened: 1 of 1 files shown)


sylvestre:
please run rustfmt

src/sed/processor.rs
```diff
@@ -306,6 +306,29 @@ fn substitute(
 
         pattern.set_to_string(result, pattern.is_newline_terminated());
 
+        // Execute the pattern space as a shell command if the 'e' flag is set
+        if sub.execute {
+            let cmd_str = pattern.as_str()?.to_string();
+            let output_bytes = std::process::Command::new("/bin/sh")
```
sylvestre: /bin/sh ? 
it won't work on windows

sylvestre:
please also add tests in tests/by-util/test_sed.rs

PranavRJoshi:
Thanks for the guidance, will work into it.

src/sed/processor.rs
PranavRJoshi: Should I make it conditionally compile on UNIX-like system only and warn out the insufficient capabilities on Windows, or do I instead execute the `cmd.exe` for Windows? The latter one would still require conditional compilation, so i can add `#[cfg]` for UNIX-like and Windows.

sylvestre: nope, it is a cross platform project. this should be a tier-1 feature on Windows too

PranavRJoshi:
Ran code coverage locally using `llvm-cov` and looked into the missing coverage for `processor.rs` file. The `map_err` error branch was the one that was uncovered based on the given changes introduced in this PR. This path is taken only when spawning the shell fails.

It's possible to cover this path too, but it requires introducing a new function which returns the name of the shell. In my opinion, this would add redundant complexity.

PranavRJoshi:
@sylvestre Do I need to make any further modifications? I think I've covered tests, but if some additional tests are required, I can work on it.

PR #415, opened 2026-05-06
add a precommit conf

sylvestre:


cakebaker:
Looks like a duplicate of https://github.com/uutils/sed/pull/414

PR #419, opened 2026-05-08
--posix should reject GNU substitute flags i/I #401

### LINKED ISSUE #401: `--posix` should reject GNU substitute flags `i`/`I`
## Bug

`--posix` does not actually reject the GNU `i` / `I` substitute flag.
The flag is silently accepted, which violates POSIX strictness when
the user opts into it.

## Reproduction

```sh
$ echo a | /usr/bin/sed --posix 's/a/b/i'
sed: -e expression #1, char 7: unknown option to 's'

$ echo a | ./target/release/sed --posix 's/a/b/i'
b                # silently honored, should be an error
```

## What it should do

When `--posix` is set, all of the GNU substitute-flag extensions
(`i`, `I`, `m`, `M`, `e`) must be rejected with the same
"unknown option to 's'" wording style we already use for unknown flags.

`--posix` is already plumbed elsewhere (e.g. `compiler.rs:1289` uses it
to choose the address-count of `=`), so the value is available.

## Suspected place to add it

`src/sed/compiler.rs:854` — `compile_subst_flags`. The function
already reaches a `format!("invalid substitute flag: '{other}'")`
branch around line 938. The fix is in the existing `'i' | 'I' =>`
arm: when `posix == true`, fall through to the same error path.

```rust
'i' | 'I' => {
    if posix {
        return compilation_error( … "unknown option to 's'" );
    }
    subst.ignore_case = true;
    line.advance();
}
```

Same treatment for `m`/`M`/`e` once those land (issues #9, #15).

`posix-mode-s` (the `for opt in i I m M ;` loop at the top of the
script).

Devel08:
resolves #401


tests/by-util/test_sed.rs
```diff
@@ -1303,3 +1303,13 @@ fn test_print_first_line_no_newline() {
         .succeeds()
         .stdout_is("foo");
 }
+
+//--posix should reject GNU substitute flags i/I https://github.com/uutils/sed/issues/401
+#[test]
```
dspinellis, line 1609: Please also add a test for the `m` flag (different code path).

src/sed/compiler.rs
```diff
@@ -899,10 +900,19 @@ pub fn compile_subst_flags(
             }
 
             'i' | 'I' => {
+                if context.posix {
+                    return compilation_error(lines, line, "unknown option to 's'");
```
dspinellis: Define a constant for the error message and use it throughout. (See current `main` branch.)

dspinellis (changes requested):
Almost there!

src/sed/compiler.rs
Devel08: But that wouldn't comply with the current code style, for example, the error message isn't a constant on line 386

tests/by-util/test_sed.rs
Devel08: done :)

src/sed/compiler.rs
dspinellis: You're almost right, but [things change](https://github.com/uutils/sed/blob/main/src/sed/compiler.rs#L386) 😃. As I wrote, check the *current* main branch.

dspinellis:
Can I review for merging? If not, please ping me when ready.

Devel08:
> Can I review for merging? If not, please ping me when ready.

It's ready for merge IMO.

dspinellis (approved):
Excellent!

PR #420, opened 2026-05-09
Implements `v` command #397

### LINKED ISSUE #397: Add `v` command (version assertion)
## Bug

The GNU `v` command — assert that the running sed is at least a given
version — is not implemented. It is essentially a no-op (succeeds if
the version is high enough, errors otherwise) and is commonly used at
the top of GNU sed scripts as a safety check.

## Reproduction

```sh
$ echo a | /usr/bin/sed 'v4.0'
a                  # accepted, prints input as usual

$ echo a | ./target/release/sed 'v4.0'
sed: <script argument 1>:1:1: error: invalid command code `v'
```

Error case (asserting a future version):

```sh
$ echo a | /usr/bin/sed 'v9.0'
sed: -e expression #1, char 4: expected newer version of sed

$ echo a | ./target/release/sed 'v9.0'
sed: <script argument 1>:1:1: error: invalid command code `v'
```

## What it should do

From the GNU manual:

> `v VERSION`  This command does nothing, but makes sed fail if GNU
> extensions are not supported, simply because other implementations
> of sed do not understand it. […] If specified, VERSION must be no
> newer than the version of GNU sed currently parsing the script;
> otherwise, the script will be aborted.

Argument is optional. When absent, the command succeeds (just an
extension assertion). When present, parse `MAJOR[.MINOR[.PATCH]]` and
compare to a constant in the source.

A reasonable choice for the "current version" constant is `4.8` or
`4.9` (matching what GNU advertises) — pick whichever lets the most
real-world scripts run unmodified, since the comparison is "no newer
than".

## Suspected place to add it

`src/sed/compiler.rs:1276` — `get_cmd_spec`. Add:

```rust
'v' => Ok(CommandSpec {
    n_addr: 0,
    handler: compile_version_command,   // new
}),
```

The handler reads the rest of the line as an optional version string
and stashes the version comparison result. Execution is a no-op
(the version check happens at compile time and just errors out then).

Rejected under `--posix`.

`compile-tests`, `stdin-prog` (uses `v9` in a probe), and any GNU
script in the wild that starts with `v`.

Devel08:
This pr implements the version command and resolves #397


dspinellis:
Thank you for the contribution! Please check the failed CI tasks.

Devel08:
@dspinellis No problem! Fixed the failing CI tasks, can you approve the workflows please?

src/sed/compiler.rs
```diff
@@ -1254,6 +1254,63 @@ fn compile_text_command_posix(
...
+fn compile_version_command(
+    lines: &mut ScriptLineProvider,
+    line: &mut ScriptCharProvider,
+    _cmd: &mut Command,
+    _context: &mut ProcessingContext,
+) -> UResult<CommandHandling> {
+    // GNU sed 4.9
+    const GNU_MAJOR: u8 = 4;
+    const GNU_MINOR: u8 = 9;
+    const GNU_PATCH: u8 = 0;
```
sylvestre: why this ?

Devel08: Because most users scripts use GNU sed 4.9, the reason is described in "What it should do" in issue #397

tests/by-util/test_sed.rs
```diff
@@ -1303,3 +1303,19 @@ fn test_print_first_line_no_newline() {
...
+        .stderr_is("sed: <script argument 1>:1:6: error: expected newer version of sed\n");
+}
+
+#[test]
+fn test_invalid_version() {
+    new_ucmd!()
+        .args(&["v4.a"])
+        .fails()
+        .stderr_is("sed: <script argument 1>:1:5: error: invalid version of sed\n");
+}
```
dspinellis, line 2267: Please also add a passing version.

src/sed/compiler.rs
dspinellis: Agreed.

tests/by-util/test_sed.rs
dspinellis: Please also add a test for the default version `v` and partial version `v4` and `v999`. See https://www.gnu.org/software/sed/manual/sed.html#Programming-Commands

src/sed/compiler.rs
```diff
@@ -1254,6 +1254,63 @@ fn compile_text_command_posix(
...
+        if line.current() == '.' {
+            ver_semantic += 1;
+            line.advance();
+        } else {
+            match ver_semantic {
+                0 => major.push(line.current()),
+                1 => minor.push(line.current()),
+                2 => patch.push(line.current()),
+                _ => return compilation_error(lines, line, "invalid version of sed"),
+            }
```
dspinellis, line 1407: Also support partial versions and defaults. So, `v` and `v4` should work correctly, defaulting to `4.0`.

dspinellis (changes requested):
Thanks!  Please address the comments, rebase the code, squash the commits, and request another review.

src/sed/compiler.rs
```diff
@@ -1356,6 +1413,10 @@ fn get_cmd_spec(
             n_addr: 2,
             handler: compile_trans_command,
         }),
+        'v' => Ok(CommandSpec {
```
dspinellis: Add a POSIX guard here.

tests/by-util/test_sed.rs
Devel08: Alright

src/sed/compiler.rs
```diff
@@ -1373,6 +1373,76 @@ fn compile_text_command_posix(
...
+        }
+    }
+
+    if major.is_empty() {
+        major = GNU_MAJOR.to_string();
+        minor = GNU_MINOR.to_string();
+        patch = GNU_PATCH.to_string();
+    }
+
+    let minor = if minor.is_empty() {
```
dspinellis: Since `minor` is mutable consider the more succinct:
```Rust
if minor.is_empty() {
    minor.push('0');
}
```

src/sed/compiler.rs
```diff
@@ -1373,6 +1373,76 @@ fn compile_text_command_posix(
...
+        "0".to_string()
+    } else {
+        minor
+    };
+    let patch = if patch.is_empty() {
```
dspinellis: Same here.

src/sed/compiler.rs
```diff
@@ -1373,6 +1373,76 @@ fn compile_text_command_posix(
...
 }
 
+// Handle v
+fn compile_version_command(
+    lines: &mut ScriptLineProvider,
+    line: &mut ScriptCharProvider,
+    _cmd: &mut Command,
+    _context: &mut ProcessingContext,
+) -> UResult<CommandHandling> {
+    // GNU sed 4.9
```
dspinellis: Better: `// Claim version partify with GNU sed 4.9`

src/sed/compiler.rs
```diff
@@ -1373,6 +1373,76 @@ fn compile_text_command_posix(
...
+                    compilation_error(lines, line, "expected newer version of sed")
+                }
+                Err(_) => compilation_error(lines, line, "invalid version of sed"),
+            },
+            Err(_) => compilation_error(lines, line, "invalid version of sed"),
+        },
+        Err(_) => compilation_error(lines, line, "invalid version of sed"),
+    }
+}
+
```
dspinellis, line 1441: You can avoid deep nesting (and make error messages clearer) thus:
```Rust
let Ok(major_int) = major.parse::<u8>() else {
    return compilation_error(lines, line, "invalid major version of sed");
};
let Ok(minor_int) = minor.parse::<u8>() else {
    return compilation_error(lines, line, "invalid minor version of sed");
};
let Ok(patch_int) = patch.parse::<u8>() else {
    return compilation_error(lines, line, "invalid patch version of sed");
};

if major_int <= GNU_MAJOR
    && minor_int <= GNU_MINOR
    && patch_int == GNU_PATCH
{
    Ok(CommandHandling::Continue)
} else {
    compilation_error(lines, line, "expected newer version of sed")
}
```

tests/by-util/test_sed.rs
```diff
@@ -2208,6 +2208,48 @@ fn test_print_first_line_no_newline() {
...
+
+#[test]
+fn test_default_version() {
+    // defaults to v4.9.0
+    new_ucmd!().args(&["v"]).succeeds();
+}
+
```
dspinellis, line 2252: Great!

dspinellis (changes requested):
A few more changes and we should be good to go! Don't forget to squash your commits please.

src/sed/compiler.rs
Devel08: Alright, don't we have to match GNU sed's error messages though?

dspinellis: AFAIK we must match output exactly and also error behavior. I don't think we need to match exactly the error message text. But if you see that this influences the GNU compatibility tests, feel free to ignore this comment.

Devel08: Actually, my concern is that it might cause trouble for scripts that look for that error message

Devel08:
@dspinellis done, can't squash the commits as there were ~60 commits, can you squash & merge?

src/sed/compiler.rs
dspinellis: I see.  Scripts should interpret the exit code, not try to parse error messages. 😃

dspinellis (approved):

dspinellis:
Thanks!

PR #424, opened 2026-05-13
Add e substitution flag#413

PranavRJoshi:
This PR introduces e substitution mentioned in #404 and supersedes #413.

## What this does
Implements the `e` substitute flag, which executes the result of the substitution as a shell command and replaces the pattern space with stdout (minus trailing newline).

```shell
$ echo a | ./target/release/sed  's/.*/echo hi/e'
hi

$ echo a | ./target/release/sed --posix 's/.*/echo hi/e'
sed: <script argument 1>:1:14: error: the 'e' substitute flag is not allowed with --posix or --sandbox
```

## Changes
- `src/sed/command.rs`: new `execute: bool` field on `Substitution`
- `src/sed/compiler.rs`: `'e'` arm in `compile_subst_flags`; rejected at compile time under `--posix` or `--sandbox`.
- `src/sed/processor.rs`: shell execution via `/bin/sh -c` after substitution

## Tests
- Three new unit tests for the flag parser covering the intended path, --posix rejection, and --sandbox rejection.
- Full `cargo test` run passes.
- GNU testsuite tests still in progress.


PranavRJoshi:
Was looking into GNU's testsuite and the test files that was mentioned in the issue. The file `subst-options.sh` does not seem to have dedicated `e` flag tests. I did manage to find some tests for `e` flag in `execute-tests.sh` file.

dspinellis (approved):

PR #427, opened 2026-05-15
fix: allow commands before closing brace

### LINKED ISSUE #392: `}` not accepted as a command terminator
## Bug

A closing brace `}` is not accepted as a command terminator. Idioms like
`{p}`, `1{p}`, `/re/{p}` fail because the parser only accepts `;` or
end-of-line after a command.

## Reproduction

```sh
$ echo a | ./target/release/sed '{p}'
sed: <script argument 1>:1:3: error: extra characters at the end of the p command

$ echo a | ./target/release/sed '{p;}'   # workaround: trailing semicolon
a
a
```

GNU sed 4.9:

```sh
$ echo a | /usr/bin/sed '{p}'
a
a
```

## Suspected fix

`src/sed/compiler.rs:514` — `parse_command_ending`:

```rust
fn parse_command_ending(
    lines: &ScriptLineProvider,
    line: &mut ScriptCharProvider,
    cmd: &mut Command,
) -> UResult<()> {
    if !line.eol() && line.current() == ';' {
        line.advance();
        return Ok(());
    }

    if !line.eol() {
        return compilation_error( … "extra characters at the end of … command" );
    }
    Ok(())
}
```

Add a third accepted terminator: `}`. Crucially, do **not** advance past
it — leave it in the stream so that the outer loop's
`compile_end_group_command` (line 1300) can consume it as the block
close. Something like:

```rust
if !line.eol() && line.current() == '}' {
    return Ok(());
}
```

before the existing `extra characters` error.

`compile-tests`, `compile-errors`, and incidental in many others that
use `{cmd}` blocks.

MukundaKatta:
## Summary
- accept `}` as a command terminator without consuming it, so the group-close handler can still process the brace
- add a regression test for `{p}`, `1{p}`, and `/a/{p}` scripts

Fixes #392.

## Tests
- `cargo fmt --check`
- `cargo test command_may_end_before_closing_brace`
- `cargo test test_sed`
- `cargo test`
- `cargo run --quiet -- '{p}' <<<'a'`

### DIFF
--- a/src/sed/compiler.rs
+++ b/src/sed/compiler.rs
@@ -520,6 +520,10 @@ fn parse_command_ending(
         return Ok(());
     }
 
+    if !line.eol() && line.current() == '}' {
+        return Ok(());
+    }
+
     if !line.eol() {
         return compilation_error(
             lines,
--- a/tests/by-util/test_sed.rs
+++ b/tests/by-util/test_sed.rs
@@ -274,6 +274,17 @@ check_output!(addr_range_odd, ["-n", "1~2p", LINES1]);
 check_output!(addr_range_step_zero, ["-n", "10~0p", LINES1]);
 check_output!(addr_range_end_multiple, ["-n", "/l1_2/,~10p", LINES1]);
 
+#[test]
+fn command_may_end_before_closing_brace() {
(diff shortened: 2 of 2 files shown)


dspinellis (approved):
Thanks!

PR #428, opened 2026-05-15
feat: add z command

### LINKED ISSUE #396: Add `z` command (clear pattern space)
## Bug

The GNU `z` command — clear the pattern space (sets it to empty
without printing) — is not implemented.

## Reproduction

```sh
$ printf "a\nb\n" | /usr/bin/sed 'z'
                 # two empty lines

$ printf "a\nb\n" | ./target/release/sed 'z'
sed: <script argument 1>:1:1: error: invalid command code `z'
```

## What it should do

From the GNU manual:

> `z`  This command empties the contents of the pattern space.

Equivalent to `s/.*//` but never sets the substitution flag (so does
not affect `t`/`T`). Takes 0 or 1 address; rejected under `--posix`.

## Suspected place to add it

```rust
'z' => Ok(CommandSpec {
    n_addr: 2,
    handler: compile_empty_command,
}),
```

…and in `processor.rs` add an execution case that does
`pattern_space.clear()` (or equivalent for the type used).

Note: not to be confused with the `-z` / `--null-data` CLI flag, which
is already implemented at `mod.rs:106`.

Direct test (`z.sh`) is not in the runner's set, but `compile-errors`
asserts that `z` is rejected under `--posix`. Once `z` exists,
`bad_command` rejection works automatically.

MukundaKatta:
## Summary
- add GNU sed's `z` command outside POSIX mode
- clear the pattern-space contents while preserving newline state for automatic printing
- cover default output, `-n` + explicit print, POSIX rejection, and substitution-flag behavior

Fixes #396.

## Tests
- `cargo fmt --check`
- `cargo test pattern_clear_with_z`
- `cargo test test_sed`
- `cargo test`
- `cargo run --quiet -- 'z' <<'RUNEOF'`
  `a`
  `b`
  `RUNEOF`

### DIFF
--- a/src/sed/compiler.rs
+++ b/src/sed/compiler.rs
@@ -1327,6 +1327,10 @@ fn get_cmd_spec(
             n_addr: 2,
             handler: compile_empty_command,
         }),
+        'z' if !posix => Ok(CommandSpec {
+            n_addr: 2,
+            handler: compile_empty_command,
+        }),
         'l' => Ok(CommandSpec {
             n_addr: 2,
             handler: compile_number_command,
--- a/src/sed/processor.rs
+++ b/src/sed/processor.rs
@@ -621,6 +621,12 @@ fn process_file(
                     let trans = extract_variant!(command, Transliteration);
                     transliterate(&mut pattern, trans)?;
                 }
+                'z' => {
+                    // Clear the pattern contents, but preserve newline state
+                    // so automatic printing still emits an empty record.
+                    let (pat_content, _) = pattern.fields_mut()?;
+                    pat_content.clear();
+                }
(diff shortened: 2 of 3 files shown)


MukundaKatta:
Pushed a test-only follow-up for the Ubuntu cargo-test failure. The POSIX rejection case no longer pipes stdin into a process that exits before reading it, avoiding the broken-pipe panic from the test harness. Local checks: `cargo fmt --check`, `cargo test pattern_clear_with_z_is_non_posix`, `cargo test test_sed`, and `git diff --check`.

dspinellis (approved):
LGTM; thanks!

PR #429, opened 2026-05-15
feat: add F command

### LINKED ISSUE #393: Add `F` command (print current input filename)
## Bug

The GNU `F` command — print the current input filename, followed by a
newline — is not implemented.

## Reproduction

```sh
$ echo abc | /usr/bin/sed -n 'F' /etc/hostname
/etc/hostname

$ echo abc | ./target/release/sed -n 'F' /etc/hostname
sed: <script argument 1>:1:1: error: invalid command code `F'
```

## What it should do

From GNU sed manual:

> `F`  Print out the file name of the current input file (with a trailing newline).

Behaviour notes:

- Output goes to stdout (like `=`, `p`).
- For stdin, GNU prints `-`.
- Honors `-z` (NUL-separated output) — see `nulldata.sh`.
- Accepts an address range like other GNU extensions.
- Rejected under `--posix` (already handled by the existing
  `bad_command(OPT)` path; once `F` is added, the `--posix` rejection
  test in `compile-errors` will keep working as expected).

## Suspected place to add it

```rust
'F' => Ok(CommandSpec {
    n_addr: 2,
    handler: compile_empty_command,   // F takes no args
}),
```

…plus a corresponding execution case in `src/sed/processor.rs` (search
for the `match cmd.code` near `processor.rs:635`). The processor
already has access to the current input filename via the same path that
`-i` uses to compute backup names — see `src/sed/in_place.rs` for how
filenames are tracked.

`follow-symlinks`, `nulldata`. Also unblocks the `F` parts of
`follow-symlinks-stdin`.

MukundaKatta:
## Summary
- add GNU sed's `F` command outside POSIX mode
- track a raw input filename separately from the quoted error-context name
- cover file input, stdin, NUL-terminated output under `-z`, and POSIX rejection

Fixes #393.

## Tests
- `cargo fmt --check`
- `cargo test pattern_print_filename_with_f_command`
- `cargo test test_sed`
- `cargo test`
- `cargo run --quiet -- -n 'F' tests/fixtures/sed/input/lines1 | head -n 3`

### DIFF
--- a/src/sed/command.rs
+++ b/src/sed/command.rs
@@ -42,6 +42,9 @@ pub struct ProcessingContext {
     // Other context
     /// Currently processed input file name (not script) in quoted form
     pub input_name: String,
+    /// Currently processed input file name as supplied, for commands that
+    /// print it directly.
+    pub raw_input_name: String,
     /// Current input line number
     pub line_number: usize,
     /// True if this is the last address of a range
--- a/src/sed/mod.rs
+++ b/src/sed/mod.rs
@@ -204,6 +204,7 @@ fn build_context(matches: &ArgMatches) -> ProcessingContext {
 
         // Other context
         input_name: "<stdin>".to_string(),
+        raw_input_name: "-".to_string(),
         line_number: 0,
         last_address: false,
         last_line: false,
--- a/src/sed/compiler.rs
+++ b/src/sed/compiler.rs
@@ -1327,6 +1327,10 @@ fn get_cmd_spec(
             n_addr: 2,
             handler: compile_empty_command,
         }),
+        'F' if !posix => Ok(CommandSpec {
+            n_addr: 2,
+            handler: compile_empty_command,
+        }),
         'l' => Ok(CommandSpec {
             n_addr: 2,
             handler: compile_number_command,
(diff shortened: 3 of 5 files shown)


src/sed/command.rs
```diff
@@ -42,6 +42,9 @@ pub struct ProcessingContext {
     // Other context
     /// Currently processed input file name (not script) in quoted form
     pub input_name: String,
+    /// Currently processed input file name as supplied, for commands that
+    /// print it directly.
+    pub raw_input_name: String,
```
dspinellis, line 47: I see no reason for a second name.

src/sed/mod.rs
```diff
@@ -204,6 +204,7 @@ fn build_context(matches: &ArgMatches) -> ProcessingContext {
 
         // Other context
         input_name: "<stdin>".to_string(),
```
dspinellis, line 206: It think this should now be `OsString` (see other examples in the code).Using `-` rather than `<stdin>` is OK.

dspinellis (changes requested):
Thanks! Please address the comments, rebase, squash, and ask for another review.

sylvestre:
please rebase it

dspinellis:
Dealing correctly with OsString and rebasing requires a very different design. Will reimplement on current HEAD.

PR #430, opened 2026-05-15
feat: add W command

### LINKED ISSUE #398: Add `W` command (write first line of pattern space)
## Bug

The GNU `W` command — write **only the first line** of the pattern
space (up to the first embedded `\n`) to a file — is not implemented.
It is the partial-write counterpart to `w`.

## Reproduction

```sh
$ printf "abc\ndef\n" | /usr/bin/sed -n 'N;W/dev/stdout'
abc

$ printf "abc\ndef\n" | ./target/release/sed -n 'N;W/dev/stdout'
sed: <script argument 1>:1:3: error: invalid command code `W'
```

## What it should do

From the GNU manual:

> `W FILENAME`  Write to the given filename the portion of the pattern
> space up to the first newline.

Same filename-parsing as `w`. Opens the file once, appends each
invocation. Honors `--sandbox` rejection (already plumbed for `w`).

## Suspected place to add it

`src/sed/compiler.rs:1276` — `get_cmd_spec`. Add:

```rust
'W' => Ok(CommandSpec {
    n_addr: 2,
    handler: compile_write_file_command,   // same parser as 'w'
}),
```

In the execution path (search for the `w` case in `processor.rs`),
write only the bytes up to the first `\n` instead of the whole pattern
space.

`missing-filename`, `sandbox` (both also test `W` rejection paths).

MukundaKatta:
## Summary
- add GNU sed's `W` command outside POSIX mode
- reuse the existing `w` filename parser/writer and write only the first pattern-space line
- cover `N;W file` behavior and POSIX rejection

Fixes #398.

## Tests
- `cargo fmt --check`
- `cargo test write_first_line_with_w_command`
- `cargo test test_sed`
- `cargo test`
- `printf 'abc\ndef\n' | cargo run --quiet -- -n 'N;W/dev/stdout'`

### DIFF
--- a/src/sed/compiler.rs
+++ b/src/sed/compiler.rs
@@ -1348,7 +1348,7 @@ fn get_cmd_spec(
             n_addr: 2,
             handler: compile_subst_command,
         }),
-        'w' => Ok(CommandSpec {
+        'w' | 'W' if !posix => Ok(CommandSpec {
             n_addr: 2,
             handler: compile_write_file_command,
         }),
--- a/src/sed/processor.rs
+++ b/src/sed/processor.rs
@@ -607,6 +607,15 @@ fn process_file(
                     let writer = extract_variant!(command, NamedWriter);
                     writer.borrow_mut().write_line(pattern.as_str()?)?;
                 }
+                'W' => {
+                    // Append only the first line of the pattern space.
+                    let writer = extract_variant!(command, NamedWriter);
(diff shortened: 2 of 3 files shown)


sylvestre:
please rebase it

dspinellis:
Following inactivity, I will issue a new corrected and rebased PR.

PR #431, opened 2026-05-15
feat: add m/M substitute flag

### LINKED ISSUE #399: Add `m`/`M` multiline substitute flag
## Bug

The `m` / `M` substitute flag (multiline mode — make `^` and `$`
match at embedded newlines inside the pattern space) is not
implemented.

## Reproduction

```sh
$ printf "foo\nbar\n" | /usr/bin/sed 'N;s/^./X/gm'
Xoo
Xar

$ printf "foo\nbar\n" | ./target/release/sed 'N;s/^./X/gm'
sed: <script argument 1>:1:11: error: invalid substitute flag: 'm'
```

Without the `m` flag, `^` only matches the start of the pattern space,
so only the first line gets `X`:

```sh
$ printf "foo\nbar\n" | /usr/bin/sed 'N;s/^./X/g'
Xoo
bar
```

## What it should do

GNU sed manual:

> `M`  `m`  This is a GNU extension. If specified, it causes `^` and
> `$` to match respectively (in addition to the normal behavior) the
> empty string after a newline, and the empty string before a newline.

In the underlying regex engine (`fancy-regex`), this corresponds to
turning on the `(?m)` (multiline) inline flag for the compiled regex.

## Suspected place to add it

`src/sed/compiler.rs:854` — `compile_subst_flags`. Around the existing
`'i' | 'I'` branch (line 891), add:

```rust
'm' | 'M' => {
    subst.multiline = true;     // new field on Subst
    line.advance();
}
```

Then in `src/sed/fast_regex.rs` (or wherever the regex is compiled
from the substitution command), prepend `(?m)` to the pattern when
`subst.multiline` is true — analogous to how `subst.ignore_case` adds
`(?i)`.

Rejected under `--posix` (handled by issue #11 once that lands).

`subst-options`, `posix-mode-s`.

MukundaKatta:
## Summary
- add `m`/`M` substitute flag parsing
- compile substitution regexes with `(?m)` when multiline mode is requested
- cover `^` and `$` matching around embedded newlines plus compiler flag/unit coverage

Fixes #399.

## Tests
- `cargo fmt --check`
- `cargo test subst_multiline_flag`
- `cargo test test_compile_re_multiline`
- `cargo test test_compile_subst_flag_m`
- `cargo test test_sed`
- `cargo test`
- `printf 'foo\nbar\n' | cargo run --quiet -- 'N;s/^./X/gm'`

### DIFF
--- a/src/sed/compiler.rs
+++ b/src/sed/compiler.rs
@@ -460,7 +460,7 @@ fn compile_address(
             }
 
             Ok(Address::Re(compile_regex(
-                lines, line, &re, context, icase,
+                lines, line, &re, context, icase, false,
             )?))
         }
         '$' => {
@@ -628,6 +628,7 @@ fn compile_regex(
     pattern: &str,
     context: &ProcessingContext,
     icase: bool,
+    multiline: bool,
 ) -> UResult<Option<Regex>> {
     if pattern.is_empty() {
         return Ok(None);
@@ -640,11 +641,17 @@ fn compile_regex(
         &bre_to_ere(pattern)
     };
 
-    // Add case-insensitive modifier if needed.
-    let pattern = if icase {
-        format!("(?i){pattern}")
-    } else {
+    let mut modifiers = String::new();
+    if icase {
+        modifiers.push('i');
+    }
+    if multiline {
+        modifiers.push('m');
+    }
+    let pattern = if modifiers.is_empty() {
         pattern.to_string()
+    } else {
+        format!("(?{modifiers}){pattern}")
     };
 
     // Compile into engine.
@@ -793,7 +800,7 @@ fn compile_subst_command(
     subst.replacement = compile_replacement(lines, line)?;
     compile_subst_flags(lines, line, &mut subst)?;
 
-    if pattern.is_empty() && subst.ignore_case {
+    if pattern.is_empty() && (subst.ignore_case || subst.multiline) {
         return compilation_error(
             lines,
             line,
@@ -801,8 +808,15 @@ fn compile_subst_command(
         );
     }
 
-    // Compile regex with now known ignore_case flag.
-    subst.regex = compile_regex(lines, line, &pattern, context, subst.ignore_case)?;
+    // Compile regex with now known modifier flags.
+    subst.regex = compile_regex(
+        lines,
+        line,
+        &pattern,
+        context,
+        subst.ignore_case,
+        subst.multiline,
+    )?;
 
     // Catch invalid group references at compile time, if possible.
     if let Some(regex) = &subst.regex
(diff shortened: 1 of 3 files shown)


src/sed/compiler.rs
```diff
@@ -903,6 +918,11 @@ pub fn compile_subst_flags(
                 line.advance();
             }
 
+            'm' | 'M' => {
```
dspinellis, line 921: This is a GNU extension, so report an error if running under strict POSIX mode.

src/sed/compiler.rs
```diff
@@ -1583,10 +1603,23 @@ mod tests {
     #[test]
     fn test_compile_re_invalid() {
         let (lines, chars) = dummy_providers();
-        let result = compile_regex(&lines, &chars, "a[d", &ctx(), false);
+        let result = compile_regex(&lines, &chars, "a[d", &ctx(), false, false);
         assert!(result.is_err()); // Should fail due to open bracketed expression
     }
 
+    #[test]
+    fn test_compile_re_multiline() {
```
dspinellis, line 1611: Please also test `foo$`.

tests/by-util/test_sed.rs
```diff
@@ -318,6 +318,24 @@ check_output!(subst_re_reuse, ["-e", r"2s//M/;1s/l/L/", LINES1]);
 check_output!(subst_newline_class, ["-n", r"1{;N;s/[\n]/X/;p;}", LINES1]);
 check_output!(subst_newline_re, ["-n", r"1{;N;s/\n/X/;p;}", LINES1]);
 
+#[test]
+fn subst_multiline_flag_matches_embedded_line_start() {
+    new_ucmd!()
+        .arg("N;s/^./X/gm")
+        .pipe_in("foo\nbar\n")
+        .succeeds()
+        .stdout_is("Xoo\nXar\n");
```
dspinellis, line 327: Nice!

dspinellis (changes requested):
Great work!  I added suggestions for a few small changes.  To finish this feature please also consider the required support for the special character sequences (\` and \') which always match the beginning or the end of the buffer.  If this works now, just add the corresponding integration tests. Otherwise, I suggest doing this in a separate PR.

dspinellis:
Hi @MukundaKatta 

Many thanks for this and the other PRs you've submitted! Over a month ago I commented on this PR, but you haven't responded. Maybe you're too busy? Please let us know if you're still interested. Otherwise, in a couple of days I'll start introducing your commits into new PRs where I'll try to fix any issues I see and someone else will review.  I'll, of course, credit you through a `Co-authored-by` Git header.

dspinellis:
Taking over due to unresponsive author.

PR #432, opened 2026-05-15
feat: add R command

### LINKED ISSUE #394: Add `R` command (read one line from file)
## Bug

The GNU `R` command (read **one** line from a file each time it is
executed, queue it for output after the current cycle) is not
implemented. It is the single-line counterpart to `r`.

## Reproduction

```sh
$ printf "x\ny\n" > a; printf "1\n2\n" > b
$ /usr/bin/sed -e '1Rb' -e '2Rb' a
x
1
y
2

$ ./target/release/sed -e '1Rb' a
sed: <script argument 1>:1:2: error: invalid command code `R'
```

## What it should do

From the GNU manual:

> `R FILENAME`  Queue a line of FILENAME to be read and inserted into
> the output stream at the end of the current cycle, or when the next
> input line is read. Note that if FILENAME cannot be read, or if its
> end is reached, no line is appended, without any error indication.

Each execution reads exactly one line (until newline) and queues it.
Reading a non-existent file is silently ignored, and reaching EOF on
subsequent calls is silently ignored.

## Suspected place to add it

`src/sed/compiler.rs:1276` — `get_cmd_spec`. There is already an `r`
command using `compile_read_file_command` at line 1333. Add:

```rust
'R' => Ok(CommandSpec {
    n_addr: 2,
    handler: compile_read_file_command,   // same handler, different exec semantics
}),
```

Then in `processor.rs`, the execution side needs to remember a per-`R`
file handle and read one line per invocation. The existing `r` command
slurps the whole file at the end of the cycle; `R` instead does
per-cycle line reads.

`cmd-R`.

MukundaKatta:
## Summary
- add GNU sed's `R` command outside POSIX mode
- share one lazy reader per filename so repeated `R` commands consume one line at a time
- silently ignore missing files and EOF, matching GNU behavior
- cover repeated reads, missing/EOF behavior, and POSIX rejection

Fixes #394.

## Tests
- `cargo fmt --check`
- `cargo test read_one_line`
- manual reproduction: `cargo run --quiet -- -e "1R<file>" -e "2R<file>" <input>`
- `cargo test test_sed`
- `cargo test`

### DIFF
--- a/src/sed/command.rs
+++ b/src/sed/command.rs
@@ -16,6 +16,8 @@ use crate::sed::script_line_provider::ScriptLineProvider;
 
 use std::cell::RefCell;
 use std::collections::HashMap;
+use std::fs::File;
+use std::io::BufReader;
 use std::path::PathBuf; // For file descriptors and equivalent
 use std::rc::Rc;
 use uucore::error::UResult;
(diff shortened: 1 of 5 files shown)


sylvestre:
please fix the conflict

dspinellis:
Following inactivity, I will issue a new corrected and rebased PR.

PR #434, opened 2026-05-15
fix: preserve latest binary executable mode

### LINKED ISSUE #277: tag/latest-commit is missing executable permission
```
fish: Unknown command. '/tmp/DL/sed-x86_64-unknown-linux-gnu' exists but is not an executable file.
```

Maybe, we should not `zstd` directly?

MukundaKatta:
## Summary
- package the latest-commit Linux binary as a `tar.zst` archive instead of a raw zstd-compressed file
- preserve the executable permission bit when users extract the archive
- update the README wording to describe the latest-commit download as an archive

Fixes #277.

## Tests
- local archive check: `tar -C <dir> -cf - sed | zstd -19 -o sed-x86_64-unknown-linux-gnu.tar.zst`, then `zstd -dc ... | tar -xf -` and `test -x sed`
- `git diff --check`
- `cargo test test_version`
- `cargo test`

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -56,7 +56,7 @@ jobs:
         ## Build Rust sed binary
         cd 'sed'
         cargo build --config=profile.release.strip=true --profile=release #-fast
-        zstd -19 target/release/sed -o ../sed-x86_64-unknown-linux-gnu.zst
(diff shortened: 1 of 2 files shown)


oech3:
There are 2 archives now. Needed to manually remove old one.

PR #439, opened 2026-05-22
regex: accept literal [ in bracket expressions

kevinburke:
POSIX allows '[' to represent itself inside a bracket expression unless it starts a character class, collating symbol, or equivalence class construct. The parser was consuming the following character after a literal '[', which made '[[]' look unterminated, and the Rust regex backend also rejects unescaped literal '[' in classes.

Reproduce the compatibility gap by comparing these commands:

    printf 'x\n' | ./target/debug/sed -E 's/[[]/X/'

    printf 'x\n' | gsed -E 's/[[]/X/'

    printf 'x\n' | ./target/debug/sed -E 's/[^[]/X/'

    printf 'x\n' | gsed -E 's/[^[]/X/'

Before this change, uutils sed rejected the scripts as unterminated or invalid regexes while GNU sed parsed them cleanly. After this change the outputs match GNU sed: x, X, x, and X for the reported cases.

Leave the following character available for normal class parsing, then escape literal '[' characters inside parsed classes before compiling with the regex backend. Add parser, compiler, and command-level regressions for the failing sed -E substitution cases.

Observed this while trying to install Python 3.14.5 with Pyenv and uutils `sed` on the PATH, which failed.

### DIFF
--- a/src/sed/compiler.rs
+++ b/src/sed/compiler.rs
@@ -628,6 +628,73 @@ fn bre_to_ere(pattern: &str) -> String {
     result
 }
 
+fn escape_literal_open_brackets_in_classes(pattern: &str) -> String {
+    let mut result = String::with_capacity(pattern.len());
+    let mut chars = pattern.chars().peekable();
+
+    while let Some(c) = chars.next() {
+        match c {
+            '\\' => {
+                result.push('\\');
+                if let Some(escaped) = chars.next() {
+                    result.push(escaped);
+                }
+                continue;
+            }
+            '[' => result.push('['),
+            _ => {
+                result.push(c);
+                continue;
+            }
+        }
+
+        if chars.peek() == Some(&'^') {
+            result.push('^');
+            chars.next();
+        }
+
+        if chars.peek() == Some(&']') {
+            result.push(']');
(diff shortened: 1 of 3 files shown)


src/sed/compiler.rs
```diff
@@ -619,6 +619,73 @@ fn bre_to_ere(pattern: &str) -> String {
     result
 }
 
+fn escape_literal_open_brackets_in_classes(pattern: &str) -> String {
```
sylvestre: please add a rustdoc comment here :)

kevinburke: Done

src/sed/compiler.rs
```diff
@@ -619,6 +619,91 @@ fn bre_to_ere(pattern: &str) -> String {
     result
 }
 
+/// Escape literal `[` characters that appear inside a bracket expression.
```
sylvestre, line 658: Well it is a bit long now :/

kevinburke: Oh, sorry, ok.

kevinburke: How is this

PR #448, opened 2026-05-29
feat: add T command (branch on no-substitution)

### LINKED ISSUE #395: Add `T` command (branch on no-substitution)
## Bug

The GNU `T` command — branch to label **if no `s///` substitution has
been performed** since the last input line was read — is not
implemented. It is the inverse of `t`.

## Reproduction

```sh
$ echo a | /usr/bin/sed -n 'Tend; p; :end'
a

$ echo a | ./target/release/sed -n 'Tend; p; :end'
sed: <script argument 1>:1:1: error: invalid command code `T'
```

## What it should do

From the GNU manual:

> `T LABEL`  If no `s///` has done a successful substitution since the
> last input line was read or branch was taken, branch to LABEL.

Same address/parsing rules as `t`. Already-implemented `t` lives in
`compile_label_command` (compiler.rs:1308); `T` should be parsed by the
same handler.

## Suspected place to add it

```rust
'b' | 't' | 'T' => Ok(CommandSpec {
    n_addr: 2,
    handler: compile_label_command,
}),
```

…then in the execution side (look near the `t` handler in
`processor.rs`), branch when the substitution flag is *false* instead of
true, and clear it the same way `t` does.

Indirect — the broader `compile-tests`/`compile-errors` suites assert
that `T` is rejected under `--posix` and accepted otherwise.

MukundaKatta:
## Summary

Implements the GNU `T LABEL` command (issue #395): branch to `LABEL` if no `s///` substitution has succeeded since the last input line was read or the last `t`/`T` was taken. It is the inverse of `t`.

Before:
```
$ echo a | ./target/release/sed -n 'Tend; p; :end'
sed: <script argument 1>:1:1: error: invalid command code `T'
```
After:
```
$ echo a | ./target/release/sed -n 'Tend; p; :end'      # no s/// -> T branches over p
$ echo a | ./target/release/sed -n 's/a/X/; Tend; p; :end'
X                                                        # s/// ran -> T does not branch
```

## Implementation

- `get_cmd_spec`: accept `T` alongside `b`/`t` (2 addresses, `compile_label_command`). Label parsing is already code-agnostic.
- `resolve_branch_targets`: include `T` when converting `Label` -> `BranchTarget`.
- Processor: `T` branches when `!substitution_made`; in both the branch and no-branch arms the flag is reset on evaluation, matching `t`.

## Testing

- `cargo test`: full suite green (282 unit + 183 integration).
- New `branch_no_sub_simple` (selective branch) and `branch_no_sub_clears` (proves `T` clears the substitution flag) integration tests, plus a `get_cmd_spec` unit test covering `b`/`t`/`T`.

Closes #395


dspinellis (changes requested):
Generally OK. T must be disabled in POSIX mode. I will fix it.

dspinellis:
Original author is not responding to change requests. Replacing this PR with #508.

PR #449, opened 2026-05-29
fix: correct \c escape handling in replacements (#403)

### LINKED ISSUE #403: `\c\\` and `\c\d` replacement escapes mis-handled
## Bug

The `\cX` control-character escape in the `s///` replacement is
mis-handled in two ways:

1. `\c\\` (escape control-char of `\`, which is `\x1c`) is mis-parsed
   as a substitute terminator and triggers an error instead of
   producing the control character.
2. `\c\d` (a backslash escape immediately after `\c`) is silently
   accepted as `d` instead of being rejected as "recursive escaping
   after `\c` not allowed".

Simple cases like `\cA` (→ `\x01`) work correctly.

## Reproduction

```sh
$ echo a | /usr/bin/sed 's/./\cA/' | od -c | head -1
0000000 001  \n         # \cA -> \x01 (Ctrl-A)

$ echo a | ./target/release/sed 's/./\cA/' | od -c | head -1
0000000 001  \n         # OK
```

```sh
$ echo a | /usr/bin/sed 's/./\c\\/' | od -c | head -1
0000000 034  \n         # \c\ -> \x1c (Ctrl-\)

$ echo a | ./target/release/sed 's/./\c\\/'
sed: :0:10: error: ...
                        # parser bails before producing the char
```

```sh
$ echo a | /usr/bin/sed '1s/./\c\d/'
sed: -e expression #1, char 10: recursive escaping after \c not allowed

$ echo a | ./target/release/sed '1s/./\c\d/'
d                       # silently produced 'd', no error
```

## What it should do

GNU semantics for `\cX` in a replacement string:

- The character following `\c` is taken raw (no further escape
  processing). Its low 5 bits are taken, XOR'd with `0x40`, producing
  the control char. So `\cA` → `\x01`, `\c\` → `\x1c`, etc.
- A backslash *escape* (`\X`) is **not** allowed after `\c`. GNU
  rejects this with the wording above. (See
  `gnu.sed/testsuite/recursive-escape-c.sh` for the exact error
  string.)

## Suspected place to add it

The `\c` handling is part of the replacement parser in
`src/sed/compiler.rs:649` (`compile_replacement`). The fix is to
look at the literal next char (don't recurse into escape parsing) and
emit `(next & 0x1f) ^ 0x40`. If the next char is `\`, accept it as
the literal `\` (giving `\x1c`); if the *char after* `\` is anything
other than `\` itself, raise the "recursive escaping" error.

`recursive-escape-c` directly. Also relevant for `normalize-text`
which exercises `\c` in `y` strings.

MukundaKatta:
## Summary

Fixes the two `\c` replacement-escape bugs in #403, verified against GNU sed 4.10.

`\cX` takes the character after `\c` raw. GNU allows exactly one exception: a backslash escaped as `\\` denotes a literal backslash, so `\c\\` yields Ctrl-`\` (0x1c). Any *other* backslash escape after `\c` is rejected.

Before:
```
$ echo a | ./target/release/sed 's/./\c\\/'      # expected \x1c
sed: :0:10: error: unterminated substitute replacement

$ echo a | ./target/release/sed '1s/./\c\d/'     # expected error
d                                                # silently produced 'd'
```
After (matches GNU sed 4.10):
```
$ echo a | ./target/release/sed 's/./\c\\/' | od -An -tx1
 1c 0a
$ echo a | ./target/release/sed '1s/./\c\d/'
sed: ...: recursive escaping after \c not allowed
```

## Implementation

- The `\c` arm of `parse_char_escape` now consumes its argument correctly: a leading `\` must be followed by another `\` (literal backslash → control char); otherwise it raises the recursive-escaping error.
- `parse_char_escape` returns `UResult<Option<char>>` so the error propagates. All five call sites (regex, replacement, transliteration, character class, GNU text commands) thread `lines` and `?`.
- Non-ASCII-after-`\c` keeps its prior literal-`c` fallback (byte-level GNU behavior is out of scope for a char-based parser).

## Testing

- `cargo test`: full suite green (285 unit + 181 integration).
- New unit tests for `\cx`, `\c\\`, `\c\d`, and trailing-backslash.
- Verified byte-for-byte against GNU sed 4.10 for `\cA`, `\c\\`, `\cx`, `\c\d`, and mixed cases.

## Out of scope

`\c` immediately followed by the closing delimiter (e.g. `s/./\c/`) is a separate, delimiter-dependent edge case (GNU emits a literal `\`); it already errored on `main` and is not part of #403.

Closes #403

### DIFF
--- a/src/sed/delimited_parser.rs
+++ b/src/sed/delimited_parser.rs
@@ -84,9 +84,14 @@ fn create_control_char(x: char) -> Option<char> {
 /// transliterarion) and return the corresponding char.
 /// At entry line.current() must have advanced after the `\\`.
 /// Advance line to the first character not part of the escape.
-/// Return `None` if an invalid escape has been specified.
-pub fn parse_char_escape(line: &mut ScriptCharProvider) -> Option<char> {
-    match line.current() {
+/// Return `Ok(None)` if no recognized escape is present (the caller then
+/// treats the sequence literally). Return `Err` for escapes that are
+/// recognized but malformed, such as recursive escaping after `\c`.
+pub fn parse_char_escape(
+    lines: &ScriptLineProvider,
+    line: &mut ScriptCharProvider,
+) -> UResult<Option<char>> {
+    let decoded = match line.current() {
         'a' => {
             line.advance();
             Some('\x07')
@@ -117,14 +122,39 @@ pub fn parse_char_escape(line: &mut ScriptCharProvider) -> Option<char> {
         }
 
         'c' => {
-            // Control character escape: \cC
+            // Control character escape: \cX. The character after \c is taken
+            // raw (no further escape processing). GNU sed allows exactly one
+            // exception: a backslash may be escaped as \\ to denote a literal
+            // backslash (so \c\\ yields Ctrl-\, i.e. 0x1c). Any other
+            // backslash escape after \c is rejected.
             line.advance(); // move past 'c'
-            match create_control_char(line.current()) {
-                Some(decoded) => {
-                    line.advance();
-                    Some(decoded)
+            if line.eol() {
+                // Nothing to control-escape; treat the lone 'c' literally.
+                Some('c')
+            } else if line.current() == '\\' {
+                line.advance(); // consume the first backslash
+                if line.eol() || line.current() != '\\' {
+                    return compilation_error(
+                        lines,
+                        line,
+                        "recursive escaping after \\c not allowed",
+                    );
                 }
-                None => Some('c'),
+                // \\ -> literal backslash as the control argument.
+                let decoded = create_control_char('\\');
+                line.advance(); // consume the second backslash
+                decoded
+            } else {
+                // Take the next char raw. If it is not control-mappable
+                // (e.g. non-ASCII), fall back to a literal 'c' without
+                // consuming the char, matching prior behavior.
+                match create_control_char(line.current()) {
+                    Some(decoded) => {
+                        line.advance();
+                        Some(decoded)
+                    }
+                    None => Some('c'),
+                }
             }
         }
 
@@ -173,7 +203,9 @@ pub fn parse_char_escape(line: &mut ScriptCharProvider) -> Option<char> {
             }
         }
         _ => None,
-    }
+    };
+
+    Ok(decoded)
 }
 
 /// Parse a POSIX RE character class returning it as a string.
@@ -275,7 +307,7 @@ fn parse_character_class(
             if line.eol() {
                 break;
             }
-            if let Some(decoded) = parse_char_escape(line) {
+            if let Some(decoded) = parse_char_escape(lines, line)? {
                 result.push(decoded);
             } else {
                 result.push('\\');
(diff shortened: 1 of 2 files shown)


src/sed/delimited_parser.rs
```diff
@@ -117,14 +122,39 @@ pub fn parse_char_escape(line: &mut ScriptCharProvider) -> Option<char> {
         }
 
         'c' => {
-            // Control character escape: \cC
+            // Control character escape: \cX. The character after \c is taken
```
dspinellis, line 125: There is considerable logic here. Consider abstracting it into a new function `parse_control_char()` in `delimited_parser.rs`, as we do with `parse_numeric_escape()`.

src/sed/delimited_parser.rs
```diff
@@ -117,14 +122,39 @@ pub fn parse_char_escape(line: &mut ScriptCharProvider) -> Option<char> {
...
             line.advance(); // move past 'c'
-            match create_control_char(line.current()) {
-                Some(decoded) => {
-                    line.advance();
-                    Some(decoded)
+            if line.eol() {
+                // Nothing to control-escape; treat the lone 'c' literally.
+                Some('c')
+            } else if line.current() == '\\' {
+                line.advance(); // consume the first backslash
```
dspinellis, line 135: Write as complete sentence: `Consume the first backslash.`

src/sed/delimited_parser.rs
```diff
@@ -117,14 +122,39 @@ pub fn parse_char_escape(line: &mut ScriptCharProvider) -> Option<char> {
...
+                if line.eol() || line.current() != '\\' {
+                    return compilation_error(
+                        lines,
+                        line,
+                        "recursive escaping after \\c not allowed",
+                    );
+                }
+                // \\ -> literal backslash as the control argument.
+                let decoded = create_control_char('\\');
+                line.advance(); // consume the second backslash
```
dspinellis, line 145: See above regarding comment.  Follow this in all comments.

src/sed/delimited_parser.rs
```diff
@@ -117,14 +122,39 @@ pub fn parse_char_escape(line: &mut ScriptCharProvider) -> Option<char> {
         }
 
         'c' => {
-            // Control character escape: \cC
+            // Control character escape: \cX. The character after \c is taken
+            // raw (no further escape processing). GNU sed allows exactly one
```
dspinellis, line 126: Please specify GNU sed version.

src/sed/delimited_parser.rs
```diff
@@ -572,6 +605,48 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_control_escape_lowercase() {
+        // \cx upcases to X then XORs 0x40 -> 0x18.
+        assert_eq!(escape_result_with_current("cx"), (Some('\x18'), None));
+    }
+
+    #[test]
+    fn test_control_escape_escaped_backslash() {
```
dspinellis, line 615: You also need to test at EOL, right?

dspinellis (changes requested):
Thanks. Please see the comments. When done please force-push the squashed commits.

sylvestre:
@MukundaKatta ping ?

sylvestre:

PR #450, opened 2026-05-29
feat: add \U \L \u \l \E case-folding in replacements (#400)

### LINKED ISSUE #400: Add `\U`/`\L`/`\u`/`\l`/`\E` case-folding in replacement
## Bug

The GNU case-folding escapes inside the `s///` replacement —
`\U` / `\L` / `\u` / `\l` / `\E` — are not supported. Today, they
are passed through literally as the letter after the backslash.

## Reproduction

```sh
$ echo abcDEF | /usr/bin/sed 's/.*/\U&/'
ABCDEF

$ echo abcDEF | ./target/release/sed 's/.*/\U&/'
UabcDEF      # \U treated as literal 'U'
```

```sh
$ echo ABCDEF | /usr/bin/sed 's/.*/\l&/'
aBCDEF

$ echo ABCDEF | ./target/release/sed 's/.*/\l&/'
\lABCDEF     # \l passed through with the backslash
```

Same issue for `\L` (all lowercase until `\E` or end), `\u` (uppercase
next char only), `\E` (end the previous `\U`/`\L`).

## What it should do

GNU semantics:

| Escape | Meaning |
| --- | --- |
| `\U` | uppercase all output until `\E` or end of replacement |
| `\u` | uppercase the very next output character |
| `\L` | lowercase all output until `\E` or end |
| `\l` | lowercase the very next output character |
| `\E` | terminate the effect of the most recent `\U`/`\L` |

These nest and apply to text from backreferences (`\1`, `&`) as well as
literal text in the replacement.

## Suspected place to add it

`src/sed/compiler.rs:649` — `compile_replacement` builds the
replacement template (the structure that `processor.rs` renders).
Add new template-element variants for each case-folding state, then
have the renderer track a small "case mode" stack while writing
output.

Look at how `\1`/`&` are handled (around `compiler.rs:1144` and the
test at `compile_replacement_backrefs_and_literal`) — that's the
pattern to mirror.

Note: the test `subst-replacement.sh` also checks that `\E` ends a
range, that `\U\1\E\2\3` only uppercases the first group, etc.

`subst-replacement`, `posix-mode-s` (where the absence of `\l` is the
expected POSIX behavior — i.e. the GNU branch must produce different
output from the POSIX branch). The fix should be gated by "not
`--posix`" so existing POSIX-mode pass-through is preserved.

MukundaKatta:
## Summary

Implements GNU case-folding escapes in the `s///` replacement (#400), verified against GNU sed 4.10.

| Escape | Meaning |
| --- | --- |
| `\U` | uppercase all output until `\E` or end |
| `\L` | lowercase all output until `\E` or end |
| `\u` | uppercase the next output character only |
| `\l` | lowercase the next output character only |
| `\E` | end the current `\U`/`\L` range |

Before:
```
$ echo abcDEF | ./target/release/sed 's/.*/\U&/'
UabcDEF      # \U treated as literal 'U'
```
After (matches GNU):
```
$ echo abcDEF | ./target/release/sed 's/.*/\U&/'
ABCDEF
```

## Implementation

- New `CaseFold` enum and `ReplacementPart::Case` variant.
- The renderer (`apply_captures` / `apply_match`) tracks a small `CaseState` (sticky range + pending one-shot) and folds text as it is emitted, so the effect spans literal text and backreferences. A one-shot (`\u`/`\l`) overrides the range for exactly one character, then the range resumes.
- Parsing is gated behind non-POSIX mode (`compile_replacement` gains a `posix` flag), preserving the existing POSIX pass-through of `\l` etc. In GNU mode these escapes take precedence over the `\u`/`\U` numeric-escape extension, matching GNU.

## Testing

- `cargo test`: full suite green (283 unit + 184 integration).
- Parser unit tests (`\U\1\E\2` parts; POSIX pass-through) and 3 integration fixtures.
- Verified byte-for-byte against GNU sed 4.10: `\U&`, `\L&`, `\u&`, `\l&`, `\U&\E suf`, `\U\1\E\2\3`, `\L\u&`, `\U\l&`, `\u\l&`, `\l\u&`, and `\Eabc`.

Closes #400


sylvestre:
needs to be rebased

PR #451, opened 2026-05-29
fix: accept a\text on the same line in POSIX mode (#144)

### LINKED ISSUE #144: `--posix -e 'a\bar'` causes an error
Another issue found with the fuzzer:
```
$ printf "foo" | cargo run -q sed --posix -e 'a\bar'
sed: <script argument 1>:1:3: error: extra characters after \ at the end of `a' command
$ printf "foo" | sed --posix -e 'a\bar'
foo
bar
```

MukundaKatta:
## Summary

Fixes the fuzzer-found bug in #144: `--posix -e 'a\bar'` errored instead of appending `bar`. Verified against GNU sed 4.10.

Before:
```
$ printf 'foo\n' | ./target/release/sed --posix -e 'a\bar'
sed: <script argument 1>:1:3: error: extra characters after \ at the end of `a' command
```
After (matches GNU):
```
$ printf 'foo\n' | ./target/release/sed --posix -e 'a\bar'
foo
bar
```

## Implementation

`compile_text_command_posix` previously ate whitespace after the `\` and rejected any remaining same-line characters. GNU instead treats the remainder of the line as the first appended text line:

- Text after `\` on the same line is taken verbatim, with **leading whitespace preserved** (`a\ bar` → ` bar`, matching GNU).
- A trailing `\` on that text escapes the newline and continues on the following line.
- `a\` with nothing after it still reports `incomplete command`.

Applies to `a`, `i`, and `c` in POSIX mode. The GNU (non-POSIX) parser was already correct and is untouched.

## Testing

- `cargo test`: full suite green (282 unit + 182 integration).
- Two prior unit tests asserted the old non-GNU rejection; they now assert the GNU-correct text content. Added a same-line-continuation unit test and a dual-mode (`check_output_posix!`) integration fixture.
- Verified byte-for-byte against GNU sed 4.10 for `a\bar`, `i\bar`, `c\bar`, leading-space preservation, continuation, and the incomplete case.

Closes #144


dspinellis (changes requested):
Thanks! I don't see [POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/) allowing text after the `a` command. Also the Seventh Edition Unix sed doesn't allow it.
```
printf 'foo\n' | ./sed -e 'a\bar'
command garbled: a\bar
```
So I think this behavior should be disabled under `--posix`.

dspinellis (changes requested):
Please fix the failing tests and squash the commits. All that has changed in the non-test code are some comments. Kindly clarify why this PR is needed.

sylvestre:

PR #458, opened 2026-06-05
fast_io.rs: replace the unsafe `libc::copy_file_range` call  by rustix

### LINKED ISSUE #442: replace unsafe libc::copy_file_range by rustix
https://github.com/uutils/sed/blob/f6e69e734bcb6fd3c1a1240cf4c67b5a983b357b/src/sed/fast_io.rs#L930-L931

Franklin-Qi:
Replace the unsafe `libc::copy_file_range` call with the safe `rustix::fs::copy_file_range` 
API to improve safety and reduce unsafe code surface.

Closes: #442

### DIFF
--- a/src/sed/fast_io.rs
+++ b/src/sed/fast_io.rs
@@ -30,6 +30,11 @@ use std::marker::PhantomData;
 use std::os::fd::RawFd;
 
 #[cfg(unix)]
+use rustix::fd::BorrowedFd;
+#[cfg(target_os = "linux")]
+use rustix::fs::copy_file_range as rustix_copy_file_range;
+
+#[cfg(unix)]
 use std::os::unix::io::AsRawFd;
 
 use std::str;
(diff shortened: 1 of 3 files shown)


src/sed/fast_io.rs
```diff
@@ -1828,6 +1835,54 @@ mod tests {
         assert_eq!(&buf, b"abcdefgh");
     }
 
+    #[test]
+    #[cfg(all(target_os = "linux", target_env = "gnu"))]
```
sylvestre: target_env isn't necessary

src/sed/fast_io.rs
```diff
@@ -927,30 +932,32 @@ fn reliable_copy_file_range(
...
-                    reliable_write(out_fd, in_ptr, pending)
-                }
-                _ => Err(err),
-            };
-        } else if ret == 0 {
-            // EOF reached
-            break;
+        let mut in_off_u64 = in_off as u64;
+        let result: std::io::Result<usize> = rustix_copy_file_range(
+            unsafe { BorrowedFd::borrow_raw(in_fd) },
```
sylvestre: can we avoid these two unsafe ?

Franklin-Qi: Thanks for the review; I'll remove those two `unsafe` blocks.

sylvestre:
@Franklin-Qi ping ? :)

sylvestre:
not happy on mac:
```

error: unused import: `std::os::unix::io::AsFd`
    --> src/sed/fast_io.rs:1053:9
     |
1053 |     use std::os::unix::io::AsFd;
     |         ^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: `-D unused-imports` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(unused_imports)]`


```

Franklin-Qi:
Thanks for the feedback; I’ve modified the compilation conditions. I don’t have a Mac here, so please take a look.

PR #459, opened 2026-06-05
fast_io.rs: use sendfile or splice if copy_file_range is not supported

### LINKED ISSUE #443: fast_io.rs: use sendfile or splice if copy_file_range is not supported
`copy_file_range` is limited e.g. `copy_file_range` to pipe is impossible. In the case, we should try extending stdout's pipe size by `fcntl` and `splice`.

Moreover, mmap should be done for memfd to prevent `SIGBUS` when input was truncated.
@dspinellis

Franklin-Qi:
- Use `splice` (via rustix) and attempt to expand the pipe size when necessary.
- Modify `reliable_copy_file_range()` to try `copy_file_range` → `splice` → `write`.
- Add Linux-based unit tests.

Closes: #443

### DIFF
--- a/src/sed/fast_io.rs
+++ b/src/sed/fast_io.rs
@@ -861,7 +861,7 @@ impl OutputBuffer {
 #[cfg(unix)]
 fn reliable_write(fd: i32, ptr: *const u8, len: usize) -> std::io::Result<usize> {
     // A thin Write-compatible wrapper around a raw file descriptor
-    // This allows us to issue and utilize the write_all implementatin.
+    // This allows us to issue and utilize the write_all implementation.
     struct FdWriter(RawFd);
 
     impl Write for FdWriter {
@@ -883,6 +883,98 @@ fn reliable_write(fd: i32, ptr: *const u8, len: usize) -> std::io::Result<usize>
     let buf: &[u8] = unsafe { std::slice::from_raw_parts(ptr, len) };
     writer.write_all(buf)?;
     Ok(len)
+}
+
+/// Wrapper to get current pipe buffer size
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+fn get_pipe_size(fd: i32) -> std::io::Result<usize> {
+    let ret = unsafe { libc::fcntl(fd, libc::F_GETPIPE_SZ) };
+    if ret < 0 {
+        Err(io::Error::last_os_error())
+    } else {
+        Ok(ret as usize)
+    }
+}
+
+/// Wrapper to set pipe buffer size
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+fn set_pipe_size(fd: i32, size: usize) -> std::io::Result<()> {
+    let ret = unsafe { libc::fcntl(fd, libc::F_SETPIPE_SZ, size as libc::c_int) };
+    if ret < 0 {
+        Err(io::Error::last_os_error())
+    } else {
+        Ok(())
+    }
+}
+
+/// Try to splice data from input file to output.
+/// Only works on Linux when both fds support splice.
+/// Return the number of bytes written, or None if splice is not supported.
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+fn try_splice(
+    in_fd: i32,
+    in_off: *mut libc::off_t,
+    out_fd: i32,
+    len: usize,
+) -> std::io::Result<Option<usize>> {
+    // Try to extend pipe buffer size if output is a pipe
+    // This helps improve throughput when writing to pipes
+    if let Ok(current_size) = get_pipe_size(out_fd) {
+        let desired_size = 1024 * 1024; // 1 MiB
+        if current_size < desired_size {
+            let _ = set_pipe_size(out_fd, desired_size);
+        }
+    }
+
+    let ret = unsafe {
+        libc::splice(
+            in_fd,
+            in_off,
+            out_fd,
+            std::ptr::null_mut(),
+            len,
+            libc::SPLICE_F_MOVE | libc::SPLICE_F_MORE,
+        )
+    };
+
+    if ret < 0 {
+        let err = io::Error::last_os_error();
+        match err.raw_os_error() {
+            Some(libc::ENOSYS) | Some(libc::EOPNOTSUPP) | Some(libc::EINVAL) => {
+                // splice not supported for these fds
+                Ok(None)
+            }
+            _ => Err(err),
+        }
+    } else {
+        Ok(Some(ret as usize))
+    }
+}
+
+/// Try to use sendfile to copy data from input to output.
+/// Only works on Linux for regular files to any output.
+/// Return the number of bytes written, or None if sendfile is not supported.
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+fn try_sendfile(
+    in_fd: i32,
+    mut in_off: libc::off_t,
+    out_fd: i32,
+    len: usize,
+) -> std::io::Result<Option<usize>> {
+    let ret = unsafe { libc::sendfile(in_fd, out_fd, &raw mut in_off, len) };
+
+    if ret < 0 {
+        let err = io::Error::last_os_error();
+        match err.raw_os_error() {
+            Some(libc::ENOSYS) | Some(libc::EOPNOTSUPP) | Some(libc::EINVAL) => {
+                // sendfile not supported for these fds
+                Ok(None)
+            }
+            _ => Err(err),
+        }
+    } else {
+        Ok(Some(ret as usize))
+    }
 }
 
 /// Copy efficiently len data from the input to the output file.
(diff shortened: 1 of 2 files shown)


oech3:
sendfile is internally splice. So no worth to try it. splice with middler pipe is faster.

src/sed/fast_io.rs
```diff
@@ -885,6 +885,98 @@ fn reliable_write(fd: i32, ptr: *const u8, len: usize) -> std::io::Result<usize>
     Ok(len)
 }
 
+/// Wrapper to get current pipe buffer size
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+fn get_pipe_size(fd: i32) -> std::io::Result<usize> {
```
oech3: No need to get size. Use set only without checking 1st size.

src/sed/fast_io.rs
```diff
@@ -885,6 +885,98 @@ fn reliable_write(fd: i32, ptr: *const u8, len: usize) -> std::io::Result<usize>
...
+        Ok(ret as usize)
+    }
+}
+
+/// Wrapper to set pipe buffer size
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+fn set_pipe_size(fd: i32, size: usize) -> std::io::Result<()> {
+    let ret = unsafe { libc::fcntl(fd, libc::F_SETPIPE_SZ, size as libc::c_int) };
+    if ret < 0 {
+        Err(io::Error::last_os_error())
```
oech3: larger pipe size is not a strong requirement. So `Err(())` is enough.

src/sed/fast_io.rs
```diff
@@ -885,6 +885,98 @@ fn reliable_write(fd: i32, ptr: *const u8, len: usize) -> std::io::Result<usize>
...
+    }
+
+    let ret = unsafe {
+        libc::splice(
+            in_fd,
+            in_off,
+            out_fd,
+            std::ptr::null_mut(),
+            len,
+            libc::SPLICE_F_MOVE | libc::SPLICE_F_MORE,
```
oech3:
```suggestion
            libc::SPLICE_F_MORE,
```
Dummy flag. Do nothing.

src/sed/fast_io.rs
```diff
@@ -2021,4 +2159,208 @@ mod tests {
...
+    ///////////////////////////////
+    // Unit tests for pipe optimization functions
+    ///////////////////////////////
+
+    #[cfg(all(target_os = "linux", target_env = "gnu"))]
+    #[test]
+    fn test_get_pipe_size_from_pipe() {
+        // Create a pipe
+        let mut fds = [0; 2];
+        assert_eq!(unsafe { libc::pipe(fds.as_mut_ptr()) }, 0);
```
oech3: `std::io::pipe`

src/sed/fast_io.rs
```diff
@@ -885,6 +885,98 @@ fn reliable_write(fd: i32, ptr: *const u8, len: usize) -> std::io::Result<usize>
...
+    len: usize,
+) -> std::io::Result<Option<usize>> {
+    // Try to extend pipe buffer size if output is a pipe
+    // This helps improve throughput when writing to pipes
+    if let Ok(current_size) = get_pipe_size(out_fd) {
+        let desired_size = 1024 * 1024; // 1 MiB
+        if current_size < desired_size {
+            let _ = set_pipe_size(out_fd, desired_size);
+        }
+    }
```
oech3:
```suggestion
    // Try to extend pipe buffer size if output is a pipe
    // This helps improve throughput when writing to pipes
    const MAX_ROOTLESS_PIPE_SIZE: usize = 1024 * 1024;
    let _ = set_pipe_size(out_fd, MAX_ROOTLESS_PIPE_SIZE);
```

src/sed/fast_io.rs
```diff
@@ -2021,4 +2159,208 @@ mod tests {
...
+        let read_fd = fds[0];
+        let write_fd = fds[1];
+
+        // Get the pipe size - should succeed and return positive value
+        let size = get_pipe_size(write_fd).expect("get_pipe_size failed");
+        assert!(size > 0, "pipe size should be positive");
+
+        // Clean up
+        unsafe {
+            libc::close(read_fd);
```
oech3: Not required if `std::io::pipe` is used

oech3:
Please use rustix instead of unsafe libc.

sylvestre:
@Franklin-Qi sorry, it needs to be rebased

src/sed/fast_io.rs
Franklin-Qi: Done — removed `get_pipe_size`; we only call `fcntl_setpipe_size` now.

src/sed/fast_io.rs
Franklin-Qi: Done — dropped `SPLICE_F_MOVE`; using `SpliceFlags::MORE` only via rustix.

src/sed/fast_io.rs
```diff
@@ -383,6 +386,7 @@ impl IOChunkContent<'_> {
 struct FastCopy {
     fd: i32,           // Raw file descriptor
     is_regular: bool,  // True if this is a regular file
+    is_fifo: bool,     // True if this is a pipe/FIFO (splice-capable)
```
oech3, line 389: We don't need to check that input or output is pipe if middler pipe is used. Also splice fails if both of in/output are not pipe. So that bool should be unnecessary.

src/sed/fast_io.rs
```diff
@@ -33,7 +33,10 @@ use std::os::fd::RawFd;
 use rustix::fd::BorrowedFd;
 #[cfg(target_os = "linux")]
 use rustix::fs::copy_file_range as rustix_copy_file_range;
-
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
```
oech3, line 36:
```suggestion
#[cfg(target_os = "linux")]
```
Same for others.

src/sed/fast_io.rs
```diff
@@ -929,6 +961,68 @@ fn reliable_write(fd: i32, ptr: *const u8, len: usize) -> std::io::Result<usize>
     Ok(len)
 }
 
+/// Best-effort rootless pipe capacity (1 MiB). Larger size is optional.
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+const MAX_ROOTLESS_PIPE_SIZE: usize = 1024 * 1024;
+
+/// Best-effort pipe capacity bump. Larger size is optional, so failures are ignored.
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+fn set_pipe_size(fd: impl AsFd, size: usize) -> Result<(), ()> {
```
oech3, line 970: Why do we need to wrap this even we just ignore the error?

src/sed/fast_io.rs
```diff
@@ -929,6 +961,68 @@ fn reliable_write(fd: i32, ptr: *const u8, len: usize) -> std::io::Result<usize>
...
+/// Only works on Linux when at least one fd is a pipe.
+/// Return the number of bytes written, or None if splice is not supported.
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+fn try_splice(
+    in_fd: BorrowedFd<'_>,
+    in_off: &mut u64,
+    out_fd: BorrowedFd<'_>,
+    len: usize,
+) -> std::io::Result<Option<usize>> {
+    match splice(in_fd, Some(in_off), out_fd, None, len, SpliceFlags::MORE) {
```
oech3, line 984: I don't think MORE is useful at here.

src/sed/fast_io.rs
```diff
@@ -929,6 +961,68 @@ fn reliable_write(fd: i32, ptr: *const u8, len: usize) -> std::io::Result<usize>
...
+            }
+        }
+    }
+}
+
+/// Copy via splice when the output is a pipe (or other splice-capable fd).
+/// Falls back to write(2) if splice is not supported.
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+fn reliable_splice(
+    in_ptr: *const u8,
```
oech3, line 1003: Why is this needed (with unsafe code)?

src/sed/fast_io.rs
```diff
@@ -1772,36 +1885,32 @@ mod tests {
...
-        let mut fds = [0; 2];
-        assert_eq!(unsafe { libc::pipe(fds.as_mut_ptr()) }, 0);
-        let read_fd = fds[0];
-        let write_fd = fds[1];
+        let (mut reader, writer) = io::pipe().unwrap();
 
         // Spawn a reader that drains slowly
         thread::spawn(move || {
             let mut buf = [0u8; 1024];
             loop {
```
oech3, line 1897: `while let Ok(1..) = reader.read(&mut buf) {...}`

oech3:
remove sendfile from PR's title and the commit message.

sylvestre:
@Franklin-Qi ping?

Franklin-Qi:
> @Franklin-Qi ping?

Yeah, I will deal with this issue soon. @sylvestre

PR #463, opened 2026-06-14
sed: fix brace quantifier delimiter and {,n} bound edge cases

sylvestre:
@LoukasPap what do you think ? :)


LoukasPap:
That's very good! Hadn't thought about simplifying the logic of `{,}` and `{,n}`, neither the delimiter case (adding the forgotten test cases was a nice catch too). I think the code is now both smarter and cleaner.

sylvestre:
thanks

PR #481, opened 2026-07-03
Implement `e` (execute) command

### LINKED ISSUE #476: Add `e` command (execute)
## Bug

The GNU `e` command — execute a shell command and send its output to the
pattern space or output stream — is not implemented. This is the standalone
command counterpart to the `s///e` substitute flag.

## Reproduction

```sh
$ echo a | /usr/bin/sed 'e echo hi'
hi
a

$ echo a | ./target/release/sed 'e echo hi'
sed: <script argument 1>:1:1: error: invalid command code `e'

Pattern-space execution form:

$ echo 'echo hi' | /usr/bin/sed 'e'
hi

$ echo 'echo hi' |  ./target/release/sed 'e'
sed: <script argument 1>:1:1: error: invalid command code `e'
```

Remember that, following the GNU documentation, the pattern space is only replaced with the bare `e` command.

```sh
$ echo a | /usr/bin/sed -e 'e echo hi' -e 's/hi/ho/'
hi
a
$ echo 'echo hi' | /usr/bin/sed -e e -e 's/hi/ho/'
ho
```

Regarding implementation, for `e` with arguments, rather than reading the shell command's input,
it's probably more efficient and desirable to flush the *sed* output and simply have the shell's output appear on the standard output.
Given the cost of invoking a shell, the cost of the flush is negligible, and, if the invoked command's output is large, the savings in CPU instructions and memory will be substantial.

Note however, that this is not how GNU sed works (notice the two *write* system calls):
```sh
$ echo a | strace -fe write /usr/bin/sed -e 'e echo hi' -e 's/hi/ho/' >/dev/null
strace: Process 3630943 attached
[pid 3630943] write(1, "hi\n", 3)       = 3
[pid 3630943] +++ exited with 0 +++
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=3630943, si_uid=1001, si_status=0, si_utime=0, si_stime=0} ---
write(1, "hi\na\n", 5)                  = 5
+++ exited with 0 +++
```

## What it should do

GNU sed manual:

> e [COMMAND] This command allows one to pipe input from a shell command
> into pattern space. Without parameters, the e command executes the command
> that is found in pattern space and replaces the pattern space with the output;
> a trailing newline is suppressed.

Behavior notes:

- With an argument, execute that argument as a shell command and (this is the default) have stdout
  appear on the output stream.

- Without an argument, execute the current pattern space as a shell command,
replace pattern space with stdout, and suppress one trailing newline.

- Accepts addresses like other commands.
- Must be rejected under --sandbox.
- Must be rejected under --posix.

## Suspected place to add it

src/sed/compiler.rs — get_cmd_spec. Add an e entry with a handler
that parses the rest of the line as an optional shell command:

'e' => Ok(CommandSpec {
n_addr: 2,
handler: compile_execute_command,
}),

The execution side belongs in src/sed/processor.rs: spawn a shell
(/bin/sh -c <command> on Unix), collect stdout, and either write it to
stdout for e COMMAND or replace pattern space for bare e.
Arrange to reuse the functionality added in b14272c1d73.

## Security note

The standalone e command is one of the commands disabled by --sandbox
along with r/w and s///e. Enforce this at compile time so the script
is rejected before any input is read.

eval, sandbox.

PranavRJoshi:
Fixes #476.

Implements the standalone `e` command (a GNU extension). Supports the command with and without arguments.

Changes:

- Added the `e` command in compiler.rs/processor.rs, both forms (with argument, bare/pattern-space). Argument parsing reuses the same escape decoding and backslash continuation as a/c/i, as GNU sed's `e` argument follows the same syntax.
- Rejected under `--posix` and `--sandbox` at compile time.
- Fixed `e` and `s///e` to silently swallowing the child shell's stderr.
- Added 17 tests covering both command forms, addressing, ';' not separating the argument, escape decoding, backslash continuation, --posix/--sandbox rejection, and stderr surfacing on failure.

Irrelevant: fixed a typo found at `delimited_parser.rs`.

### DIFF
--- a/src/sed/processor.rs
+++ b/src/sed/processor.rs
@@ -223,7 +223,7 @@ fn shell_stdout(cmd: &str) -> io::Result<Vec<u8>> {
     let mut child = shell_command(cmd)
         .stdout(std::process::Stdio::piped())
         .spawn()?;
-    let mut stdout = child.stdout.take().expect("stdout was piped");
+    let mut stdout = child.stdout.take().expect("stdout should be piped");
     let mut buf = Vec::new();
     stdout.read_to_end(&mut buf)?;
     child.wait()?;
@@ -232,7 +232,7 @@ fn shell_stdout(cmd: &str) -> io::Result<Vec<u8>> {
 
 /// Execute the pattern space as a shell command, replacing its contents
 /// with the command's standard output, minus one trailing newline.
-fn execute_pattern_as_shell(
+fn execute_pattern_as_shell_command(
     pattern: &mut IOChunk,
     command: &Command,
     context: &mut ProcessingContext,
@@ -246,12 +246,23 @@ fn execute_pattern_as_shell(
         )
         .unwrap_err()
     })?;
-    let mut shell_out = String::from_utf8_lossy(&stdout_bytes).into_owned();
-    if shell_out.ends_with("\r\n") {
-        // On windows, both return carriage and newline characters are used
-        shell_out.truncate(shell_out.len() - 2);
-    } else if shell_out.ends_with('\n') {
-        // Strip the trailing newline, as GNU sed does
+    let mut shell_out = String::from_utf8(stdout_bytes).map_err(|e| {
+        input_runtime_error::<()>(
+            &command.location,
+            context,
+            format!("shell command output is not valid UTF-8: {e}"),
+        )
+        .unwrap_err()
+    })?;
+    #[cfg(windows)]
+    {
+        if shell_out.ends_with("\r\n") {
+            // On Windows a trailing \r\n is the line terminator; strip both.
+            shell_out.truncate(shell_out.len() - 2);
+        }
+    }
+    // Unix (and some Windows tool) has the trailing \n. Strip it, as GNU sed does.
+    if shell_out.ends_with('\n') {
         shell_out.pop();
     }
     pattern.set_to_string(shell_out, pattern.is_newline_terminated());
@@ -371,7 +382,7 @@ fn substitute(
 
         // Execute the pattern space as a shell command if the 'e' flag is set
         if sub.execute {
-            execute_pattern_as_shell(pattern, command, context)?;
+            execute_pattern_as_shell_command(pattern, command, context)?;
         }
 
         if sub.print_flag {
@@ -609,7 +620,7 @@ fn process_file(
                 }
                 'e' => match &command.data {
                     CommandData::None => {
-                        execute_pattern_as_shell(&mut pattern, &command, context)?;
+                        execute_pattern_as_shell_command(&mut pattern, &command, context)?;
                     }
                     CommandData::Text(cmd_str) => {
                         let stdout_bytes = shell_stdout(cmd_str).map_err(|e| {
@@ -620,7 +631,15 @@ fn process_file(
                             )
                             .unwrap_err()
                         })?;
-                        output.write_str(String::from_utf8_lossy(&stdout_bytes).into_owned())?;
+                        let shell_out = String::from_utf8(stdout_bytes).map_err(|e| {
+                            input_runtime_error::<()>(
+                                &command.location,
+                                context,
+                                format!("shell command output is not valid UTF-8: {e}"),
+                            )
+                            .unwrap_err()
+                        })?;
+                        output.write_str(shell_out)?;
                     }
                     _ => panic!("invalid 'e' command data"),
                 },
--- a/src/sed/processor.rs
+++ b/src/sed/processor.rs
@@ -19,7 +19,7 @@ use crate::sed::named_writer;
 
 use std::borrow::Cow;
 use std::cell::RefCell;
-use std::io::{self, IsTerminal};
+use std::io::{self, IsTerminal, Read};
 use std::path::PathBuf;
 use std::rc::Rc;
 use uucore::display::Quotable;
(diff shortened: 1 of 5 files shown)


PranavRJoshi:
Seems like CRLF issue on Windows. Will make those tests conditionally compile on UNIX only.

src/sed/processor.rs
```diff
@@ -215,6 +215,49 @@ fn shell_command(_cmd: &str) -> std::process::Command {
...
 
+/// Run `cmd` in a shell, returning its standard output. The child's
+/// standard error is left connected to this process's own, matching GNU
+/// sed's behavior of letting shell errors surface directly rather than
+/// being silently captured.
+fn shell_stdout(cmd: &str) -> io::Result<Vec<u8>> {
+    let mut child = shell_command(cmd)
+        .stdout(std::process::Stdio::piped())
+        .spawn()?;
+    let mut stdout = child.stdout.take().expect("stdout was piped");
```
dspinellis: Better: "stdout should be piped"

src/sed/processor.rs
```diff
@@ -215,6 +215,49 @@ fn shell_command(_cmd: &str) -> std::process::Command {
...
+    let stdout_bytes = shell_stdout(&cmd_str).map_err(|e| {
+        input_runtime_error::<()>(
+            &command.location,
+            context,
+            format!("failed to execute shell command: {e}"),
+        )
+        .unwrap_err()
+    })?;
+    let mut shell_out = String::from_utf8_lossy(&stdout_bytes).into_owned();
+    if shell_out.ends_with("\r\n") {
```
dspinellis: Make this active only for `#[cfg(windows)]`. I wanted to comment on the previous PR, but let it pass to move ahead.

src/sed/processor.rs
```diff
@@ -215,6 +215,49 @@ fn shell_command(_cmd: &str) -> std::process::Command {
...
+        // On windows, both return carriage and newline characters are used
+        shell_out.truncate(shell_out.len() - 2);
+    } else if shell_out.ends_with('\n') {
```
dspinellis: But keep this also for Windows, because some Windows tools use \n.

src/sed/processor.rs
```diff
@@ -215,6 +215,49 @@ fn shell_command(_cmd: &str) -> std::process::Command {
...
+    let cmd_str = pattern.as_str()?.to_string();
+    let stdout_bytes = shell_stdout(&cmd_str).map_err(|e| {
+        input_runtime_error::<()>(
+            &command.location,
+            context,
+            format!("failed to execute shell command: {e}"),
+        )
+        .unwrap_err()
+    })?;
+    let mut shell_out = String::from_utf8_lossy(&stdout_bytes).into_owned();
```
dspinellis: As *sed* is a data processing tool rather than a UI front-end, we should avoid `from_utf8_lossy`, which can lead to silent data corruption. I suggest using (through a suitable method) `IOChunk::from_content`.

src/sed/processor.rs
```diff
@@ -328,24 +371,7 @@ fn substitute(
...
-            let mut shell_out = String::from_utf8_lossy(&output_bytes.stdout).into_owned();
-            if shell_out.ends_with("\r\n") {
-                // On windows, both return carriage and newline characters are used
-                shell_out.truncate(shell_out.len() - 2);
-            } else if shell_out.ends_with('\n') {
-                // Strip the trailing newline, as GNU sed does
-                shell_out.pop();
-            }
-            pattern.set_to_string(shell_out, pattern.is_newline_terminated());
+            execute_pattern_as_shell(pattern, command, context)?;
```
dspinellis: Great reuse!

src/sed/processor.rs
```diff
@@ -215,6 +215,49 @@ fn shell_command(_cmd: &str) -> std::process::Command {
...
+    let mut stdout = child.stdout.take().expect("stdout was piped");
+    let mut buf = Vec::new();
+    stdout.read_to_end(&mut buf)?;
+    child.wait()?;
+    Ok(buf)
+}
+
+/// Execute the pattern space as a shell command, replacing its contents
+/// with the command's standard output, minus one trailing newline.
+fn execute_pattern_as_shell(
```
dspinellis: Better names `execute_pattern_through_shell` or `execute_pattern_as_shell_command` .

src/sed/processor.rs
```diff
@@ -581,6 +607,23 @@ fn process_file(
...
+                    CommandData::Text(cmd_str) => {
+                        let stdout_bytes = shell_stdout(cmd_str).map_err(|e| {
+                            input_runtime_error::<()>(
+                                &command.location,
+                                context,
+                                format!("failed to execute shell command: {e}"),
+                            )
+                            .unwrap_err()
+                        })?;
+                        output.write_str(String::from_utf8_lossy(&stdout_bytes).into_owned())?;
```
dspinellis: Please avoid `from_utf8_lossy`

tests/by-util/test_sed.rs
```diff
@@ -632,6 +632,217 @@ fn test_subst_e_flag_no_match_no_exec() {
         .stdout_is("hello\n");
 }
 
+////////////////////////////////////////////////////////////
+// e command (execute)
+#[cfg(unix)]
```
dspinellis, line 795: Add a separate corresponding test for Windows.

dspinellis (changes requested):
Thank you for the quick turnaround and thorough testing!  I added a few improvement suggestions.

PranavRJoshi:
Appreciate the thorough review. Will make the appropriate changes and commit them later.

dspinellis (changes requested):
I see some potential clashes with the highly relevant #487. Are you OK to delay the review a bit until we can land that so that we can ensure compatibility? In the meantime, please
* [ ] try to address the drop in code coverage (I thought it would be addressed by your integration tests) , and
* [ ] see why the GNU test suite results haven't changed. You can run the suite through `util/run-gnu-testsuite.sh`.

PranavRJoshi:
I went through the scripts exercising `e`/`s///e`. One of them, `eval.sh`, required strict flag ordering of `p` and `e`, which has been added. The other three that didn't pass was:

- `sandbox.sh`: The rejection behavior is correct. But it fails on the format, i.e., our sed reports in the format `<script argument N>:L:C: error:` while GNU sed reports in the format `-e expression #1, char N:`.
- `posix-mode-s.sh`: same format issue as well as unimplemented `\l\L\U\u\E` escapes. `s///e` execution passes.
- `execute-tests.sh`: unimplemented `W`/`T`/`F` and big-number `q`. The `e` part passes.

dspinellis (approved):
All comments have been well-addressed and the changes have been rebased (thanks).

PR #483, opened 2026-07-05
Optimize multiline matching

dspinellis:
The multiline matching resulted in a large performance regression due to the indiscriminate use of the dotall (?s) flag.  Change the regex selection process to use only the flag when probably needed.


sylvestre:
"optimize" was right ;)

PR #487, opened 2026-07-07
Improve GNU compatibility through byte processing

dspinellis:
The original Rust implementation assummed a UTF-8 environment, as is common in modern systems.  This commit improves GNU sed compatibility by changing most processing to run on bytes rather than Rust's Unicode characters and strings. This change also improves the handling of binary files. The commit adds a simple POSIX-compatible locale configuration to support UTF-8 processing when specified. While at it, also fix script-specified file handling to take place through OsString.

The change fixes two failing GNU test suite cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,12 +116,19 @@ cargo test
   sequences.
 
 ### Incompatibilities
-* The input is assumed to be valid UTF-8 (this includes 7-bit ASCII).
-  If the input is in another code page, consider converting it through UTF-8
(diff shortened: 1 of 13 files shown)


dspinellis:
I force-pushed a few unit tests to also pass `codecov/patch`.

dspinellis:
@sylvestre ideally, I'd prefer reviewing #481 _after_ we land this. But please tell me if reviewing this PR is going to take long (I realize it's big and I apologize). In that case I'll review #481 and deal with the integration clashes later.

PR #531, opened 2026-08-19
Implement GNU W (write pattern first line) command

dspinellis:
While at it, also fix the w command on patterns missing a newline.

### DIFF
--- a/src/sed/compiler.rs
+++ b/src/sed/compiler.rs
@@ -1648,6 +1648,10 @@ fn get_cmd_spec(
             n_addr: 2,
             handler: compile_label_command,
         }),
+        'W' if !posix => Ok(CommandSpec {
+            n_addr: 2,
+            handler: compile_write_file_command,
+        }),
         'w' => Ok(CommandSpec {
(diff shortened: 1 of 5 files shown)


src/sed/compiler.rs
```diff
@@ -1648,6 +1648,10 @@ fn get_cmd_spec(
             n_addr: 2,
             handler: compile_label_command,
         }),
+        'W' if !posix => Ok(CommandSpec {
```
sylvestre, line 1651: Would it make sense to use an enum at some point?

dspinellis: You mean an enum for the compatibility target, like POSIX, GNU (default), or uutils? This is an interesting alternative `--uutil-extensions` flag introduced in #537. What other values would this have?

sylvestre: oh, no, just in the code :)

dspinellis: Thanks! I think I understand. I added #538 as a proposal.

PR #534, opened 2026-08-20
fix(deps): refactor sha2 package configuration

xtqqczze:


sylvestre:
I am not sure this qualifies as a refactor

PR #535, opened 2026-08-21
Add tests for non-pattern-print on deletion

dspinellis:
These are the tests from https://github.com/uutils/sed/pull/256, which seems to have been fixed.


cakebaker:
Thanks!

PR #544, opened 2026-08-30
Implement GNU R (read one line from file) command

### LINKED ISSUE #394: Add `R` command (read one line from file)

Jorge-Polanco-Roque:
Closes #394.

Implements the GNU `R` command, the line-oriented counterpart to `r` (mirroring how `W` relates to `w`, added in #531).

`R filename` queues the *next* successive line of `filename` to be inserted into the output stream at the end of the current cycle. Each cycle consumes one more line; once the file is exhausted no line is queued.

### Behavior
```console
$ printf 'a\nb\nc\nd\n' | sed -e 'R lines.txt'   # lines.txt = x1,x2,x3
a
x1
b
x2
c
x3
d
```
- The file is opened lazily on first use and read one line at a time (new `NamedReader`, analogous to `NamedWriter`).
- An unreadable or exhausted file yields no further lines — no error is raised, matching GNU sed.
- Reserved to non-POSIX mode and rejected under `--sandbox`, exactly like `r`/`w`/`W`.

### Tests
- Unit tests for `NamedReader` (successive lines, final line without newline, missing file).
- Integration tests: successive-line behavior, silent missing file, `--posix` rejection, `--sandbox` rejection.

`cargo fmt --all -- --check`, `cargo test --all`, and `cargo clippy --all-targets --workspace -psed -- -D warnings` all pass locally.


sylvestre:
please add tests to improve the coverage. some tests are missing to reach important cases

Jorge-Polanco-Roque:
Thanks — added tests to cover the missing cases: unit tests for `compile_read_line_command` (sandbox rejection and that it installs a `NamedReader`), plus integration tests for `R` appending under `-n` (suppressed auto-print) and for a source file shorter than the input. The `R` compile and execute paths are now exercised directly.

Jorge-Polanco-Roque:
Added integration tests for R covering the cases that were missing: address-gated R that only advances when it runs, independent cursors across different files, and a final line without a trailing newline. While writing them I also caught and fixed a real divergence from GNU: multiple `R` on the same file now share one cursor instead of each restarting from line 1. Full suite and clippy are green.

Jorge-Polanco-Roque:
Gentle ping — this has been open ~10 days. I added the integration tests you asked for (the missing R cases: multiple files, address ranges, and end-of-input handling), and the GNU sed testsuite comparison is clean. Happy to rebase or adjust if anything else is needed. Thanks!

src/sed/compiler.rs
```diff
@@ -1122,6 +1123,29 @@ fn compile_read_file_command(
     Ok(CommandHandling::Continue)
 }
 
+// Handles R
```
sylvestre, line 1126: does it make a test of the GNU testsuite pass? if so, please mention it in the commit message

src/sed/compiler.rs
```diff
@@ -1122,6 +1123,29 @@ fn compile_read_file_command(
...
+    lines: &mut ScriptLineProvider,
+    line: &mut ScriptCharProvider,
+    cmd: &mut Command,
+    context: &mut ProcessingContext,
+) -> UResult<CommandHandling> {
+    if context.sandbox {
+        return compilation_error(lines, line, ERR_SANDBOX);
+    }
+    let path = read_file_path(lines, line)?;
+    // GNU sed shares a single cursor per file path among all `R` commands
```
sylvestre: the why is already documented on `named_readers` in command.rs, please keep only one :)

src/sed/compiler.rs
```diff
@@ -1636,6 +1660,10 @@ fn get_cmd_spec(
             n_addr: 2,
             handler: compile_empty_command,
         }),
+        'R' if !posix => Ok(CommandSpec {
```
sylvestre, line 1662: please also add `R` to the list of GNU extensions in README.md (next to `W`)

src/sed/named_reader.rs
```diff
@@ -0,0 +1,101 @@
+// An abstraction for input files read one line at a time by the `R` command
```
sylvestre, line 1: please squash the 5 commits into one, with the usual `sed: ` prefix

Jorge-Polanco-Roque:
Thanks @sylvestre! Addressed all four:

- Squashed the 5 commits into one with the `sed: ` prefix.
- It does make the GNU testsuite `cmd-R` test pass — mentioned that in the commit message now.
- Dropped the duplicated comment; the "why" now lives only on `named_readers` in command.rs.
- Added `R` to the GNU extensions list in the README, next to `W`.

Rebased on main while at it; fmt/tests green.

PR #545, opened 2026-08-30
sed: support back-references in byte mode

### LINKED ISSUE #543: Back-references are rejected in byte (C locale) mode
Found by the differential fuzzer (`fuzz/fuzz_targets/fuzz_sed.rs`).

In the C/POSIX locale we select the byte-oriented engine, which rejects
back-references outright. GNU sed supports them in every locale.

```console
$ printf 'aba\n' | LC_ALL=C sed -n '/\(a\)b\1/p'
aba
$ printf 'aba\n' | LC_ALL=C ./target/release/sed -n '/\(a\)b\1/p'
sed: <script argument 1>:1:11: error: invalid regex '(a)b(?:\1)': back-references are not supported in byte mode
# Exit code: 1
```

The same script works in a UTF-8 locale, so this only affects byte mode:

```console
$ printf 'aba\n' | LC_ALL=en_US.UTF-8 ./target/release/sed -n '/\(a\)b\1/p'
aba
```

The error comes from `src/sed/fast_regex.rs`. Until this is fixed the fuzzer
skips scripts using back-references; see `KNOWN_ISSUES` in
`fuzz/fuzz_targets/fuzz_sed.rs`.

lKolabrodl:
Closes #543.

## Problem

Byte mode rejected every regular expression containing a back-reference before attempting a match. This made address and substitution expressions such as `\(a\)b\1` fail whenever the `C` locale selected byte-oriented matching.

## User impact

Scripts using back-references could work in a UTF-8 locale but fail immediately in the `C` locale. Invalid UTF-8 input could not use back-references at all.

## Root cause

The fast byte regex engine does not implement back-references, while the fallback `fancy_regex` engine accepts only UTF-8 text and reports UTF-8 offsets.

## Fix

- add a byte-mode `fancy_regex` path that maps each input byte to one Unicode scalar
- translate matches and capture offsets back to the original byte buffer
- keep `captures_iter` streaming instead of collecting every match
- preserve captures from invalid UTF-8 input
- cover the reported address expression and an invalid UTF-8 substitution

## Verification

- `cargo test --lib` — 360 passed
- `cargo test backref` — 9 related tests passed
- `cargo clippy --lib --tests -- -D warnings`
- `cargo fmt --all -- --check`

The full integration run on Windows reached 219 passing tests, 2 ignored tests, and 47 failures caused by existing CRLF-sensitive shell-script and fixture expectations. The focused integration tests for this change pass.

## AI disclosure

AI assistance was used to help trace the byte/UTF-8 boundary, implement the offset mapping, and draft tests. The resulting code was reviewed and verified with the commands above.


sylvestre:
could you please add more tests? the code coverage isn't ideal:
https://app.codecov.io/gh/uutils/sed/pull/545?dropdown=coverage&src=pr&el=h1

lKolabrodl:
Added five focused unit tests in 51aaf56 covering non-ASCII byte offset mapping, optional capture groups, multiple invalid UTF-8 matches, zero-width iteration, and no-match paths. Locally, `cargo fmt --all -- --check`, `cargo test fancy_byte --lib` (7 passed), and `cargo test backref` (9 passed) all pass. Thanks!

keith-hall:
Hi, not sure if you are aware, but fancy-regex implemented some bytes-mode related input handling in https://github.com/fancy-regex/fancy-regex/releases/tag/0.19.0, maybe that can be leveraged to keep the code here more maintainable. If it doesn't suit, let me know, maybe we can improve something

PR #550, opened 2026-09-05
Fix --posix to disable the GNU regular expression escapes

### LINKED ISSUE #542: --posix should disable the GNU regular expression escapes (\? \+ \| \w \s)
Found by the differential fuzzer (`fuzz/fuzz_targets/fuzz_sed.rs`).

With `--posix`, GNU sed drops the GNU regular expression escapes: `\?`, `\+`,
`\|` are not operators and `\w`, `\s` are not classes, they all stand for the
plain character. We keep honoring them.

```console
$ printf 'a\nb\n0?\n' | sed --posix -e '/a\+/d'
a
b
0?
$ printf 'a\nb\n0?\n' | ./target/release/sed --posix -e '/a\+/d'
b
0?
```

Same for the others:

| script | GNU `--posix` deletes | we delete |
| --- | --- | --- |
| `/0\?/d` | `0?` | every line (empty match) |
| `/a\+/d` | nothing | `a` |
| `/a\|b/d` | nothing | `a` and `b` |
| `/\w/d` | nothing | every line |

Until this is fixed the fuzzer skips these scenarios; see `KNOWN_ISSUES` in
`fuzz/fuzz_targets/fuzz_sed.rs`.

LoukasPap:
Fixes #542 

Under `--posix`, GNU sed treats the GNU regular expression escapes as the literal character that follows the backslash. These are the operators `\?`, `\+` and `\|`, the character classes `\w`, `\W`, `\s` and `\S`, and the empty-width matches `\b`, `\B`, `\<`, `\>`, `` \` `` and `\'`.

`parse_regex_for_mode()` now takes a POSIX parameter and emits the literal character for these escapes. The check comes before `parse_char_escape()`, otherwise `\b` is decoded as a backspace.

### BRE and ERE

The two modes need the opposite spelling of the same literal:

- In BRE the character is emitted bare, because `bre_to_ere()` escapes the ERE metacharacters afterwards. Emitting `\?` would make `bre_to_ere()` turn it back into the operator.
- In ERE `\?`, `\+` and `\|` keep their backslash, since a backslash is already what makes them literals there. Emitting them bare would make them operators.

The remaining escapes are letters and punctuation with no meaning of their own, so they are emitted bare in both modes.

### DIFF
--- a/src/sed/delimited_parser.rs
+++ b/src/sed/delimited_parser.rs
@@ -359,15 +359,27 @@ pub fn parse_regex(
     line: &mut ScriptCharProvider,
     regex_mode: RegexMode,
 ) -> UResult<Vec<u8>> {
-    parse_regex_for_mode(lines, line, regex_mode, CharacterMode::Utf8)
+    parse_regex_for_mode(lines, line, regex_mode, CharacterMode::Utf8, false)
+}
+
+/// Return true if a backslash followed by `c` is a GNU regular expression
+/// extension, which under --posix stands for the literal character `c`.
+fn is_gnu_regex_escape(c: char) -> bool {
+    matches!(
+        c,
+        '?' | '+' | '|' | 'w' | 'W' | 's' | 'S' | 'b' | 'B' | '<' | '>' | '`' | '\''
+    )
 }
 
 /// Parse a regular expression according to the current character mode.
+/// Under `posix` the GNU regular expression extensions are disabled, with each
+/// one standing for the literal character that follows its backslash.
 pub fn parse_regex_for_mode(
     lines: &ScriptLineProvider,
     line: &mut ScriptCharProvider,
     regex_mode: RegexMode,
     character_mode: CharacterMode,
+    posix: bool,
 ) -> UResult<Vec<u8>> {
     let delimiter = scan_delimiter(lines, line)?;
     let mut result = Vec::new();
@@ -385,6 +397,17 @@ pub fn parse_regex_for_mode(
                 }
                 if line.current() == delimiter {
                     // Push escaped delimiter
+                    result.push(line.current_byte());
+                    line.advance();
+                    continue;
+                }
+                if posix && is_gnu_regex_escape(line.current()) {
+                    // ? + | must keep their backslash to remain literals.
+                    if matches!(regex_mode, RegexMode::Extended)
+                        && matches!(line.current(), '?' | '+' | '|')
+                    {
+                        result.push(b'\\');
+                    }
                     result.push(line.current_byte());
                     line.advance();
                     continue;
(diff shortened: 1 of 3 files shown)


LoukasPap:
@sylvestre While working on this I noticed that `\b`, `\<`, `\>`, `` \` `` and `\'` are also broken without `--posix`, where they never match:
```bash
$ printf 'cat catalog\n' | sed 's/\<cat\>/X/g'
X catalog
$ printf 'cat catalog\n' | ./target/debug/sed 's/\<cat\>/X/g'
cat catalog
```

The fix looks small, and differs per escape. All three engines support `\<`, `\>`, `\b`, `\A` and `\z` natively, so no lookarounds are needed:

- `\<` and `\>` already reach `Regex::new()` unchanged. They break only because `NEEDS_RE` in `fast_regex.rs` does not list them, so the pattern is taken for a literal string and `remove_escapes()` drops the backslash, leaving `<cat>`. Widening `\\[AzBb]` to `\\[AzBb<>]` is enough.
- `\b` is decoded as a backspace by `parse_char_escape()`, so it has to be intercepted before that call and emitted unchanged.
- `` \` `` and `\'` have no equivalent in the Rust engines and need translating to `\A` and `\z`.


This is the separate problem from this PR, so I left it out. Do you want to add it in this PR or open a separate issue for it?

src/sed/delimited_parser.rs
```diff
@@ -389,6 +401,17 @@ pub fn parse_regex_for_mode(
                     line.advance();
                     continue;
                 }
+                if posix && is_gnu_regex_escape(line.current()) {
+                    // ? + | must keep their backslash to remain literals.
```
sylvestre, line 405: in ERE the fallback below already emits `\?` as is, could you just skip this branch for ? + | there instead of pushing the backslash again?

tests/by-util/test_sed.rs
```diff
@@ -808,6 +808,73 @@ fn test_subst_e_flag_no_match_no_exec() {
...
+    new_ucmd!()
+        .args(&["-E", "-e", r"/\w/d"])
+        .pipe_in("aw\nx\n")
+        .succeeds()
+        .stdout_is("");
+}
+
+#[test]
+fn test_posix_keeps_posix_regex_constructs() {
+    // Groups, quantifiers, operators and character escapes are POSIX,
```
sylvestre, line 867: `\x61` isn't POSIX, does GNU sed really keep it under --posix? please double check :)

LoukasPap: Verified, sed keeps it under `--posix`. But the test name doesn't capture the case so I'll move it in a new one including `\o141` and `\d097`.

PR #552, opened 2026-09-08
l command: honor the -l wrap width

### LINKED ISSUE #551: l command: the wrap width comes from the terminal size, and -l is ignored
The wrap width used by the `l` command does not come from the `-l` option or from its documented default. It comes from the width of the terminal the process happens to be running in, and falls back to 60 when there is no terminal. The same command on the same input therefore produces different bytes depending on where the output goes, and changes when the window is resized.

Tested with uutils sed 0.2.0 built from `main` at 77f442d, against GNU sed 4.9, under `LC_ALL=C`.

Measured under a pty of an explicit width, input 200 `x`, length of the first output line:

| terminal columns | 20 | 40 | 80 | 120 | 200 |
|---|---|---|---|---|---|
| GNU sed 4.9 | 70 | 70 | 70 | 70 | 70 |
| uutils sed | 20 | 40 | 80 | 120 | 200 |

`-l N` is parsed and then ignored:

```console
$ printf 'aaaaaaaaaa\n' | sed -n -l 5 l
aaaa\
aaaa\
aa$

$ printf 'aaaaaaaaaa\n' | ./target/release/sed -n -l 5 l
aaaaaaaaaa$
```

`-l 5`, `-l 0` and `-l 200` all produce the same output.

`l 0` folds after every character:

```console
$ echo abc | sed -n 'l 0'
abc$

$ echo abc | ./target/release/sed -n 'l 0'
\
a\
b\
c$
```

With no terminal, the default width is 60 rather than 70:

```console
$ printf 'a%.0s' {1..70} | sed -n l
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\
a$

$ printf 'a%.0s' {1..70} | ./target/release/sed -n l
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\
aaaaaaaaaaa$
```

The GNU sed manual, on `-l N`: "A length of 0 (zero) means to never wrap long lines. If not specified, it is taken to be 70." On the `l n` argument: "a length of 0 (zero) means to never wrap long lines. If omitted, the default as specified on the command line is used."

`compile_number_command` in `src/sed/compiler.rs` stores `output_width()` when `l` has no numeric argument, and `output_width()` reads `terminal_size()`. The value from `-l` is already computed correctly into `ProcessingContext::length` in `src/sed/mod.rs` (defaulting to 70) and is never read.

I have a fix for this and will open a PR.

SichenLiang:
Fixes #551.

Out of scope and unchanged here: how `l` renders a newline inside the pattern space, and `--posix` accepting the `l n` argument. GNU also derives the default width from a `COLS` environment variable; its manual does not document that, so this implements the documented default of 70.


SichenLiang:
On the CodSpeed regression: no benchmark in the suite uses `l` or passes `-l`. The eleven scripts are `""` (x2), `s///` (x4), `/re/d` (x2), `y///`, `a`, and `s/\r$//`. `number_fix` runs two `s///g` substitutions, so the code this PR changes — the `l` arm of `compile_number_command`, `ListLine`, and `write_partial_str` — is not on its execution path.

Measured locally with both binaries built the same way, `number_fix` runs at 0.97x main's time, `access_log_subst` 0.98x, `no_op` 1.01x, `y///` 1.00x.

The report also carries a "different runtime environments detected" warning.

src/sed/processor.rs
```diff
@@ -506,29 +506,66 @@ fn readable_char(ch: char) -> Cow<'static, str> {
     }
 }
 
+/// Bound on the rendered list output buffered when never wrapping.
+// A wrap width of zero never folds, so without periodic draining the whole
+// rendered line would be held in memory.  A few KB amortize the write calls
+// while keeping the buffer bounded.
+const LIST_FLUSH_THRESHOLD: usize = 4 * 1024;
```
sylvestre, line 513: do we really need the draining? the pattern space is already whole in memory, the rendered line is at most a few times that.
without it we could drop write_partial_str, never_wrap and half of the tests, no?

src/sed/processor.rs
```diff
@@ -506,29 +506,66 @@ fn readable_char(ch: char) -> Cow<'static, str> {
...
+
 /// Buffered state for rendering one list command output line.
 struct ListLine {
     buffer: String,
+    // Bytes of rendered output buffered since the line was last folded or
+    // drained; when wrapping, also the current column.
     width: usize,
-    max_width: usize,
+    // The wrap width.  An item that would make width reach it folds the
```
sylvestre: nobody will read such long comments :) please shorten them (same for new() and write_item)

src/sed/processor.rs
```diff
@@ -506,29 +506,66 @@ fn readable_char(ch: char) -> Cow<'static, str> {
...
+    /// drain the accumulated output before an item that would make it
+    /// reach the flush threshold.
     fn write_item(&mut self, output: &mut OutputBuffer, out_str: &str) -> UResult<()> {
         let out_len = out_str.len();
-        if self.width + out_len + 1 > self.max_width {
-            self.buffer.push_str("\\\n");
-            output.write_str(std::mem::take(&mut self.buffer))?;
+        if self.width + out_len + 1 > self.fold_at {
+            if self.never_wrap {
+                // Drain before appending out_str rather than after it.
```
sylvestre: 4 lines of comment for a clear(), please make it shorter

tests/by-util/test_sed.rs
```diff
@@ -1930,6 +1930,190 @@ fn list_invalid_utf8_byte_locale() {
...
+
+/// A wrap length of one folds before every character, so the first fold
+/// comes before any output has been emitted for the line.
+#[test]
+fn list_length_option_one() {
+    new_ucmd!()
+        .args(&["-n", "-l", "1", "l"])
+        .pipe_in(b"abc\n".to_vec())
+        .succeeds()
+        .stdout_is_bytes(b"\\\na\\\nb\\\nc$\n");
```
sylvestre, line 2002: a `\` before any character looks odd, could you please check what GNU sed does with -l 1? i think it doesn't wrap at all

src/sed/mod.rs
```diff
@@ -107,7 +107,9 @@ pub fn uu_app() -> Command {
                 .num_args(0..=1)
                 .default_missing_value(""),
             // Access with .get_one::<u32>("line-length")
-            arg!(-l --length <NUM> "Specify the 'l' command line-wrap length.")
+            arg!(-l --"line-length" <NUM> "Specify the 'l' command line-wrap length.")
+                // The long name used before GNU sed's --line-length was accepted.
+                .alias("length")
```
sylvestre, line 112: do we need to keep the --length alias? it isn't a GNU option

SichenLiang:
Thanks for the review. I checked each point below.

The periodic drain makes a substantial difference for `-l 0`. These are median peak-RSS values from three runs of `sed -n -l 0 l FILE`, with one input line per file:

| input line       | with draining | without | GNU sed 4.10 |
| ---------------- | ------------: | ------: | -----------: |
| 200 MB of `a`    |        204 MB |  604 MB |       402 MB |
| 200 MB of `\001` |        204 MB | 1804 MB |       402 MB |

A byte can expand to four bytes as `\001`, and retaining the growing rendered buffer accounts for the 1.8 GB peak. GNU remains at 402 MB for both inputs. I did not measure a consistent time difference: on the escaped input the two builds came out at 14.7 s and 15.2 s of CPU time, which is inside the run-to-run spread I see on this machine.

Reproducer:

```console
python3 -c 'open("esc.txt","wb").write(b"\001"*200_000_000+b"\n")'
/usr/bin/time -f '%M KB' ./target/release/sed -n -l 0 l esc.txt > /dev/null
```

Without the drain, four unit tests specific to that behavior would be removed. The 14 `l` integration tests still pass because the output is unchanged.

To reduce the added surface, I will inline `write_partial_str` at its only call site and store the width itself instead of the `never_wrap` flag. That keeps the memory bound with one less piece of state and without adding a general output-buffer method.

I will also shorten the comments in `new()`, `write_item()`, and above `clear()`.

I rechecked `-l 1`; GNU does wrap there:

```console
$ printf 'abc\n' | sed -n -l 1 l | od -c
0000000   \  \n   a   \  \n   b   \  \n   c   $  \n
```

The leading backslash and newline form an empty first segment. `-l 2` produces `a\`, `b\`, `c$`, while `-l 0` disables wrapping. I compared widths 0 through 5 on plain and escaped input, and the output matches GNU 4.9 and 4.10 byte for byte.

For `--length`, that is the name the published crates expose: 0.1.1 and 0.2.0 both declare `arg!(-l --length <NUM> ...)`. It is undocumented and untested, and GNU accepts only `--line-length`. I kept `--length` as a hidden alias to avoid breaking existing invocations, but I can remove it if undocumented option names are not considered part of the compatibility surface.

SichenLiang:
I tried inlining `write_partial_str`, but it depends on three private parts of `OutputBuffer`: `WriteRange`, `flush_mmap`, and `out`. Moving the code into `ListLine::write_item` would require exposing all three.

Using `write_str` would also change the output because it defers trailing newlines. Each 4 KiB drain would then introduce a line break; for example, 6000 `\001` bytes with `-l 0` become six lines instead of one. The `write_partial_str_defers_no_newline` test covers this distinction.

I made `write_partial_str` `pub(crate)` instead. Since the method is new in this PR, this leaves the crate's public API unchanged. I also removed `never_wrap`, use the stored width directly, and shortened the comments.

I compared the updated version with the previous build across widths 0–70, several input types, the 4 KiB boundaries, and the file, stdin, mmap, and `-i` paths. The output was identical, and peak RSS for the 100 MB escaped-line case was unchanged.

Removing `write_partial_str` entirely would also mean removing the periodic drain, so I kept the current version to preserve the memory bound.

PR #557, opened 2026-09-15
sed: implement \\U \\L \\u \\l \\E case-conversion escapes in s///

### LINKED ISSUE #540: s/// replacement: implement the \U \L \u \l \E case-conversion escapes
GNU sed supports case-conversion escapes in the replacement text of s///. None of them work in uutils sed today.

```sh
$ echo "abc def" | sed 's/b\(.*\)/\U&/'
```

| script | uutils sed | GNU sed 4.9 |
| --- | --- | --- |
| `s/b\(.*\)/\U&/` | `aUbc def` | `aBC DEF` |
| `s/b\(.*\)/\u&/` | `aubc def` | `aBc def` |
| `s/b\(.*\)/\L&/` | `a\Lbc def` | `abc def` |
| `s/b\(.*\)/\l&/` | `a\lbc def` | `abc def` |
| `s/b\(.*\)/\E&/` | `a\Ebc def` | `abc def` |

dedsec-terminal:
Fixes #540.

Implements GNU `s///` replacement case-conversion escapes:

- `\U` uppercase until `\L`/`\E`, `\L` lowercase until `\U`/`\E`
- `\u`/`\l` one-shot next-character conversion
- `\E` ends persistent conversion and clears pending one-shot

Details:
- New `ReplacementPart::{Upper,Lower,UpperFirst,LowerFirst,End}` in `command.rs`; `compile_replacement` parses them before `parse_char_escape` so `\u`/`\U` in replacements are case-conversion directives rather than `\uXXXX`/`\UXXXXXXXX` Unicode escapes (matches GNU sed behavior; README updated accordingly). Escaped delimiter still wins (e.g. `sU...U\UU`).
- Runtime `append_with_case` applies persistent + one-shot to literals, `&`, `\1`..`\9`. Empty matches leave one-shot pending (GNU `s/(b?)-/\u\1x/g` carry), any produced char consumes it. Fresh state per substituted occurrence so `g` does not propagate.
- Factored out case selection into `take_case(single, persistent)` helper.
- UTF-8 mode is Unicode-aware with invalid-byte passthrough; byte mode is ASCII-only.

Verified against GNU sed 4.9 including the issue table (`ABC DEF`, `ABc def`) plus `\U&\E`, `\U\1-\2`, empty-group `g` cases, `\u\l`/`\l\u` last-wins, `\U\l`/`\L\u` combos.

Tests: `cargo test --lib` 385 passed, `cargo test --test tests` 268 passed, `cargo clippy --lib` clean, `cargo fmt --check` clean.


dedsec-terminal:
Hi @dspinellis — friendly review request on this s/// case-conversion implementation (fixes #540; GNU suite +1, utf8-ru). Just pushed a perf follow-up (fast paths when no conversion directives; 265 tests green, fmt+clippy clean) to address the Codspeed delta. Let me know if you need anything else. Thanks!

dedsec-terminal:
Follow-up pushed (23b7709): patch coverage for the fast paths - unit tests for unmatched-group/single-literal fast paths, flag-sync contract panics, LowerFirst invalid-byte handling, compiler case-directive parsing + flag, and end-to-end substitution tests; plus removal of two dead None arms preempted by the all-None early return. Full suite green (lib + integration), fmt/clippy clean. Remaining reds are not from this change: Android jobs fail in sdkmanager setup, and CodSpeed shows -3.75% on number_fix only (genome/access_log improved; likely noise, happy to dig if you want).

.github/workflows/GnuComment.yml
```diff
@@ -12,6 +12,7 @@ jobs:
     permissions:
       actions: read  #  to list workflow runs artifacts
       pull-requests: write  #  to comment on pr
+      issues: write  #  to update/delete existing bot comments
```
sylvestre: unrelated to the case-conversion change, please split that into its own pr

src/sed/compiler.rs
```diff
@@ -771,6 +771,48 @@ pub fn compile_replacement(
                             line.advance();
                         }
 
+                        // GNU case-conversion escapes: \U \L \u \l \E.
+                        // These produce no output themselves; they control how
+                        // following replacement text (literals, &, \1..\9) is
+                        // cased. Handled before parse_char_escape so that \u
+                        // and \U are not mistaken for \uXXXX / \UXXXXXXXX
+                        // Unicode escapes (GNU reserves them for conversion).
+                        'U' => {
```
sylvestre: same 5 lines five times, could be a single arm mapping the char to the part, no?

src/sed/compiler.rs
```diff
@@ -771,6 +771,48 @@ pub fn compile_replacement(
...
+                            line.advance();
+                        }
+                        'L' => {
+                            if !literal.is_empty() {
+                                parts.push(ReplacementPart::Literal(std::mem::take(&mut literal)));
+                            }
+                            parts.push(ReplacementPart::Lower);
+                            line.advance();
+                        }
+                        'u' => {
```
sylvestre: README says `\uXXXX` / `\UXXXXXXXX` work in the replacement, this drops that silently.
please update the README (and mention it in the PR description)

src/sed/command.rs
```diff
@@ -140,6 +146,232 @@ impl Default for ReplacementTemplate {
...
+        }
+        if !wrote {
+            // to_lowercase() can yield nothing for some chars; preserve original.
+            let mut buf = [0u8; 4];
+            result.extend_from_slice(ch.encode_utf8(&mut buf).as_bytes());
+        }
+    }
+}
+
+/// Append `input` to `result` applying GNU `s///` case conversion.
```
sylvestre: nobody will read such a long comment :)

src/sed/command.rs
```diff
@@ -140,6 +146,232 @@ impl Default for ReplacementTemplate {
...
+///
+/// `persistent` selects the active `\U`/`\L` conversion, `single` holds a
+/// pending `\u`/`\l` one-shot conversion for the next produced character.
+/// An empty `input` (e.g. an unmatched group) leaves `single` untouched so
+/// it carries over to the following text, matching GNU sed. A present
+/// character — even one without case (digit, punctuation, space) — consumes
+/// `single`. Invalid UTF-8 bytes are passed through and consume `single`.
+#[inline]
+fn append_with_case(
```
sylvestre, line 200: the "which case to apply" match is written 3 times in this function, please factor it out. and the comments are quite long :)

.github/workflows/GnuComment.yml
dedsec-terminal: Reverted the workflow changes so this PR stays strictly scoped to the case-conversion escapes.

src/sed/compiler.rs
dedsec-terminal: Consolidated the five arms into a single `case @ ('U' | 'L' | 'u' | 'l' | 'E')` arm mapping to the corresponding `ReplacementPart`.

src/sed/compiler.rs
dedsec-terminal: Updated the README to clarify that in substitution replacements, `\u` and `\U` are GNU case-conversion directives, and updated the PR description accordingly.

src/sed/command.rs
dedsec-terminal: Shortened and streamlined the doc comments.

src/sed/command.rs
dedsec-terminal: Factored out the case resolution logic into a `take_case(single, persistent)` helper function and shortened the surrounding comments.

src/sed/command.rs
```diff
@@ -140,6 +146,191 @@ impl Default for ReplacementTemplate {
...
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum PersistentCase {
+    None,
+    Upper,
+    Lower,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum SingleCase {
```
sylvestre: `PersistentCase` and `SingleCase` are the same enum, one type + Option would do, no?

src/sed/command.rs
```diff
@@ -140,6 +146,191 @@ impl Default for ReplacementTemplate {
...
+            PersistentCase::Upper => Some(true),
+            PersistentCase::Lower => Some(false),
+            PersistentCase::None => None,
+        },
+    }
+}
+
+/// Decode a single UTF-8 character (or invalid byte) at the start of `bytes`.
+/// Returns the decoded char (None for invalid) and its byte length (1 for invalid).
+fn decode_one_utf8(bytes: &[u8]) -> (Option<char>, usize) {
```
sylvestre: do we really need a hand rolled utf8 decoder? `str::from_utf8` error gives you valid_up_to/error_len for free.

src/sed/command.rs
```diff
@@ -140,6 +146,191 @@ impl Default for ReplacementTemplate {
...
+        return;
+    }
+    if upper {
+        for c in ch.to_uppercase() {
+            let mut buf = [0u8; 4];
+            result.extend_from_slice(c.encode_utf8(&mut buf).as_bytes());
+        }
+    } else {
+        // lower
+        let mut wrote = false;
```
sylvestre: `to_lowercase()` always yields at least one char, so this fallback is dead code, please drop it.

src/sed/command.rs
```diff
@@ -140,6 +146,191 @@ impl Default for ReplacementTemplate {
...
+                None => result.push(b),
+            }
+        }
+        return;
+    }
+    // Validate once before the Unicode-aware fast path.
+    if let Ok(s) = std::str::from_utf8(input) {
+        let mut chars = s.chars();
+        if let Some(first_ch) = chars.next() {
+            let upper = take_case(single, persistent).expect("case conversion is active");
```
sylvestre: please avoid the expect() here, return None and copy the char instead.

src/sed/command.rs
```diff
@@ -178,42 +396,148 @@ impl ReplacementTemplate {
...
     /// Apply the template to the given RE single match.
-    pub fn apply_match(&self, m: &Match) -> Vec<u8> {
+    pub fn apply_match(&self, m: &Match, character_mode: CharacterMode) -> Vec<u8> {
         let mut result = Vec::new();
+        self.apply_match_to(m, character_mode, &mut result);
+        result
+    }
+
+    /// Appends the rendered template directly into `result` to avoid intermediate buffer allocations.
+    pub fn apply_match_to(&self, m: &Match, character_mode: CharacterMode, result: &mut Vec<u8>) {
```
sylvestre: the same match over parts is now written 4 times (fast/slow x captures/match), could you please factor it out?

src/sed/processor.rs
```diff
@@ -305,8 +306,8 @@ fn substitute(
                 Ok(Some(m)) => {
                     result.extend_from_slice(&text[last_end..m.start()]);
 
-                    let replacement = sub.replacement.apply_match(&m);
-                    result.extend_from_slice(&replacement);
+                    sub.replacement
+                        .apply_match_to(&m, character_mode, &mut result);
```
sylvestre: the apply_*_to buffer refactor is unrelated to the case escapes, please split it into its own pr — also, is the codspeed failure coming from it?

src/sed/command.rs
dedsec-terminal: Merged `PersistentCase` and `SingleCase` into a single `CaseConversion` enum and simplified `take_case` to `single.take().or(persistent)`.

src/sed/command.rs
dedsec-terminal: Dropped the custom `decode_one_utf8` helper and switched to `std::str::from_utf8` inspecting `valid_up_to()` and `error_len()` on invalid sequences.

src/sed/command.rs
dedsec-terminal: Dropped the dead code fallback after `to_lowercase()`.

src/sed/command.rs
dedsec-terminal: Replaced `.expect()` with a `push_char` helper that gracefully copies the original character when no case conversion is active.

src/sed/command.rs
dedsec-terminal: Factored out all four parts matching loops into a unified `render_parts` helper in `ReplacementTemplate`.

src/sed/processor.rs
dedsec-terminal: Reverted the buffer refactor in `processor.rs` back to `apply_match` and `apply_captures`, keeping this PR strictly focused on the case escapes. Regarding CodSpeed: no, the earlier failure was an environmental -2.9% variance on `number_fix` on an earlier commit, not related to the refactor. On the latest commit all CodSpeed benchmarks are passing (+10.36% efficiency overall).

src/sed/command.rs
```diff
@@ -531,6 +758,494 @@ mod tests {
...
+        let template = ReplacementTemplate::new(vec![
+            ReplacementPart::Upper,
+            ReplacementPart::Group(1),
+            ReplacementPart::Literal(b"-".to_vec()),
+            ReplacementPart::Group(2),
+        ]);
+        let input = &mut IOChunk::new_from_str("ab");
+        let caps = caps_for("(a)(b)", input);
+        let cmd = Command::default();
+        let out = template
```
sylvestre: a test that pokes a pub field to hit a debug_assert isn't testing much :) better to keep the flag private and make the state unrepresentable, then both wildcard arms and these two tests go away.

